PR #69, opened 2014-10-06
Enable using proprietary codecs

zcbenz:
Fixes https://github.com/atom/atom-shell/issues/633.

This PR makes Chromium recognize proprietary formats and use the right codecs if they are available. This does not build proprietary codecs in ffmpeg.


ali:
It doesn't look like this change added FLAC support to Electron. I'm still unable to decode/play FLAC files using the Web Audio API or `<audio>` elements.

PR #94, opened 2015-03-03
Upgrade to Chrome 41

zcbenz:
WIP


zcbenz:
@aroben I believe something is wrong with the `libchromiumcontent-win` CI machine, it now reports an error for both `chrome41` and `master` branch when running `gyp_chromium`:

```
No such file or directory: '...\\out\\Debug\\gyp-win-tool'
```

However yesterday the exact same commit can build without problem, do you know if there is any change about Python or gyp environment on the CI machine recently?

aroben:
Hm, have you tried deleting the workspace? Maybe there's some build artifact getting in the way.

PR #97, opened 2015-03-29
:memo: Update required VS2013 version on Windows.

hokein:
Currently, the minimal required version of VS2013 for building Chromium is Update 4. See https://www.chromium.org/developers/how-tos/build-instructions-windows

As far as I know, VS2013 Update 2 cannot build Chrome41.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ TODO
 
 #### Windows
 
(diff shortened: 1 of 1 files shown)


zcbenz:
Thanks!

PR #98, opened 2015-04-03
Do not build libchromiumcontent as a single shared library

zcbenz:
This PR makes libchromiumcontent build with 2 versions: one is `shared_library` mode, which is the "component build" in Chromium and consists lots of small shared libraries; the other is `static_library` mode, which consists static libraries that can be linked to the final binary.

The `shared_library` version is used for development, which links very fast but in sacrifice of size and performance. And the `static_library` version is used for release, it generates the binary with smallest size and maximum optimization, but in the price of extremely long link time.

In atom-shell, we will use `shared_library` version in Debug build, and  use `static_library` version in Release build.

This also enables 64bit build on Windows.


aroben:
Yay! Thanks for making this happen.

PR #100, opened 2015-04-13
Add patch to disable autohidecursorevent on mac

deepak1556:
Related https://github.com/atom/atom-shell/pull/1326


zcbenz:
It should be made optional, see how we override `acceptsFirstMouse `: 
https://github.com/atom/libchromiumcontent/blob/atom/patches/render_widget_host_view_mac.patch#L38.

zcbenz:
Patch merged in https://github.com/atom/libchromiumcontent/blob/atom/patches/render_widget_host_view_mac.patch.

PR #102, opened 2015-04-21
Upgrade to Chrome 42

zcbenz:


aroben:
I'll bet it felt nice not to update `inside_blink.patch`!

zcbenz:
This is the shortest Chrome upgrade PR! :smiley:

PR #103, opened 2015-04-27
Add `--component` argument to specify building type.

hokein:
The building script will build `static_library` and `shared_library` at one time, which requires a lot of time. It would be better to add a switch argument allowing developers to choose the specific building type. @zcbenz PLTAL


zcbenz:
Thanks!

PR #104, opened 2015-04-30
Build pdf component for win-print API support.

hokein:


deepak1556:
@hokein i am wroking on supporting ppapi, glad that you shipped ppapi headers in this patch. But there are a few ppapi components that needs to be built (ppai_*), shall i base it on this PR or you will be adding it ?

hokein:
@deepak1556  Which ppapi components will you need? The `pdf` component already has dependency on ppapi modules (`ppapi_cpp_object` and `ppapi_internal_module`).

It may require a long time to get win-print patch merged. If you need it eariler, you can directly ship ppapi without basing on this PR, and I will update this PR later.

deepak1556:
Ah cool just needed `ppapi_internal` , didnt check `pdf.gyp` . But we are yet to confirm whether to ship pdfium as an optional shared library right ? I will just work on ppapi for now without basing on this PR. Thanks!

chromiumcontent/chromiumcontent.gyp
```diff
@@ -33,6 +33,7 @@
         '<(DEPTH)/content/content.gyp:content_app_both',
         '<(DEPTH)/content/content_shell_and_tests.gyp:content_shell_pak',
         '<(DEPTH)/net/net.gyp:net_with_v8',
+        '<(DEPTH)/pdf/pdf.gyp:pdf',
```
zcbenz: This dependency is only needed on Windows. And `pdf.gyp:pdf` is hard-coded to be `static_library`, we have to deal with it specially in component build.

hokein: The component build has been finished now. For pdf.lib static library in component build, we need to link all its dependent libraries manually. Also introducing change on atom/brightray#102.

@zcbenz please take a look.

zcbenz:
:sparkles:

PR #106, opened 2015-05-05
shipping ppapi components

deepak1556:


zcbenz:
`ppapi_proxy` and `ppapi_ipc` are hard-coded to be built as shared_library, so they need to be treated specially like #104.

hokein:
@zcbenz 
As far as I know, the `ppapi_ipc` and `ppapi_proxy` aren't hard-coded in gyp. see https://code.google.com/p/chromium/codesearch#chromium/src/ppapi/ppapi_internal.gyp&sq=package:chromium&l=81&q=%22ppapi_ipc%22&type=cs

zcbenz:
@hokein Ah I misunderstood the code, thanks!

deepak1556:
thanks @hokein

PR #115, opened 2015-05-13
Ship pdf component as a shared library

hokein:
This patch makes the pdf library build from static_library to shared_library. As a side effect, we only ship the pdf.dll, and no need to ship all the pdf related libraries.

### DIFF
--- a/patches/export_pdf.patch
+++ b/patches/export_pdf.patch
@@ -1,8 +1,8 @@
 diff --git a/pdf/pdf.cc b/pdf/pdf.cc
-index 7aba412..246cd81 100644
+index 7aba412..f57902a 100644
 --- a/pdf/pdf.cc
 +++ b/pdf/pdf.cc
-@@ -81,21 +81,26 @@ const void* PPP_GetInterface(const char* interface_name) {
+@@ -81,7 +81,13 @@ const void* PPP_GetInterface(const char* interface_name) {
    return pp::Module::Get()->GetPluginInterface(interface_name);
  }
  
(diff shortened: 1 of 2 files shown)


patches/export_pdf.patch
```diff
@@ -0,0 +1,131 @@
...
+-                                 int bounds_origin_x,
+-                                 int bounds_origin_y,
+-                                 int bounds_width,
+-                                 int bounds_height,
+-                                 bool fit_to_bounds,
+-                                 bool stretch_to_bounds,
+-                                 bool keep_aspect_ratio,
+-                                 bool center_in_bounds,
+-                                 bool autorotate) {
++PDF_EXPORT bool RenderPDFPageToDC(const void* pdf_buffer,
```
zcbenz: I think `PDF_EXPORT` is more suitable to be put in the `pdf.h`?

And we can also put `PDF_EXPORT` in a separate line to make the patch minimum like this:

```c++
PDF_EXPORT
bool RenderPDFPageToDC(...
```

patches/export_pdf.patch
```diff
@@ -0,0 +1,131 @@
...
+diff --git a/pdf/pdf.gyp b/pdf/pdf.gyp
+index f47e33a..6e0a12d 100644
+--- a/pdf/pdf.gyp
++++ b/pdf/pdf.gyp
+@@ -6,7 +6,7 @@
+   'targets': [
+     {
+       'target_name': 'pdf',
+-      'type': 'static_library',
++      'type': 'loadable_module',
```
zcbenz: We can change the `type` and `defines` in `chromiumcontent.gypi`, it is preferred not to introduce a new patch if we don't have to, because maintaining patches is very painful when doing upgrade.

patches/export_pdf.patch
```diff
@@ -0,0 +1,131 @@
...
++++ b/pdf/pdf.h
+@@ -5,6 +5,22 @@
+ #ifndef PDF_PDF_H_
+ #define PDF_PDF_H_
+ 
++#if defined(WIN32)
++
++#if defined(PDF_IMPLEMENTATION)
++#define PDF_EXPORT __declspec(dllexport)
++#define PDF_EXPORT_PRIVATE __declspec(dllexport)
```
zcbenz: The `PDF_EXPORT_PRIVATE` doesn't seem to be needed.

zcbenz:
@hokein On making `pdf.dll` optional, I think we can just put it in `DelayLoadDLLs`, so we don't have to move the APIs out of the `chrome_pdf` namespace and do dynamic library loading in https://github.com/atom/electron/pull/1669. I want to make the copied source files from Chromium to have least differences from upstream, so future upgrading won't be __too__ painful.

patches/export_pdf.patch
hokein: There is a strange issue in changing the `pdf` target's `type` and `defines` in `chromiumcontent.gypi`. It doesn't link the dependent libraries while linking `pdf.dll`, which causes a link error. 

I add the following setting code in `chromiumcontent.gypi` to custom `pdf` component: 
```
      ['_target_name=="pdf"', {
        'type': 'loadable_module',
        'defines': [
          'PDF_IMPLEMENTATION',
        ],
      }],
```

patches/export_pdf.patch
hokein: Done.

patches/export_pdf.patch
hokein: > I think PDF_EXPORT is more suitable to be put in the pdf.h?

The reason for it is that to make the patch minimum. I have tried it before, it will lead to more change in `pdf.h`.

hokein:
@zcbenz 
I have minimized the `export_pdf.patch` now.

Personally, I prefer do dynamic library loading rather than `DelayLoadDLLs`.  In `DelayLoadDlls`, we need to do error handling(such as `missing the dll`) within try-catch mechanisom(https://msdn.microsoft.com/en-us/library/3aeywt27.aspx), which violates Chromium's code style.

Besides Chromium's `base` library has already provided a similar function [`LoadNativeLibrary`](https://code.google.com/p/chromium/codesearch#chromium/src/base/native_library_win.cc&q=LoadNativeLibrary&sq=package:chromium&type=cs&l=57) in making dynamic library easily. So I think the dynamic loading way is more aligned with Chromium style.

zcbenz:
> Personally, I prefer do dynamic library loading rather than DelayLoadDLLs. In DelayLoadDlls, we need to do error handling(such as missing the dll) within try-catch mechanisom(https://msdn.microsoft.com/en-us/library/3aeywt27.aspx), which violates Chromium's code style.

It sounds reasonable to me.

patches/export_pdf.patch
zcbenz: It seems that the linking stage of `pdf` still treats `pdf` as `static_library`, it is good to patch it then.

patches/export_pdf.patch
```diff
@@ -0,0 +1,97 @@
...
+--- a/pdf/pdf.h
++++ b/pdf/pdf.h
+@@ -5,6 +5,20 @@
+ #ifndef PDF_PDF_H_
+ #define PDF_PDF_H_
+ 
++#if defined(WIN32)
++
++#if defined(PDF_IMPLEMENTATION)
++#define PDF_EXPORT __declspec(dllexport)
```
zcbenz: Since `PDF_EXPORT` is only used in `pdf.cc` and it is only compiled on Windows, I think we can use `__declspec(dllexport)` directly instead of defining `PDF_EXPORT`?

hokein: We need to define `PDF_EXPORT` macro  since `pdf.cc` will be compiled on other platforms.
But we can only define `__declspec(dllexport)` here  to keep minimum change in `pdf.h`.
The patch has been updated now.

zcbenz:
:+1:

zcbenz:
@hokein The component build failed when linking `pdf.dll`:

```
[8/10] LINK_EMBED(DLL) pdf.dll
FAILED: C:\Python27\python.exe gyp-win-tool link-with-manifests environment.x64 True pdf.dll "C:\Python27\python.exe gyp-win-tool link-wrapper environment.x64 False link.exe /nologo /IMPLIB:pdf.dll.lib /DLL /OUT:pdf.dll @pdf.dll.rsp" 2 mt.exe rc.exe "obj\pdf\pdf.pdf.dll.intermediate.manifest" obj\pdf\pdf.pdf.dll.generated.manifest
pdf.pdfium_engine.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) int __cdecl printing::ConvertUnit(int,int,int)" (__imp_?ConvertUnit@printing@@YAHHHH@Z) referenced in function "public: virtual bool __cdecl chrome_pdf::PDFiumEngine::GetPageSizeAndUniformity(class pp::Size *)" (?GetPageSizeAndUniformity@PDFiumEngine@chrome_pdf@@UEAA_NPEAVSize@pp@@@Z)
pdf.pdfium_engine.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) double __cdecl printing::ConvertUnitDouble(double,double,double)" (__imp_?ConvertUnitDouble@printing@@YANNNN@Z) referenced in function "public: virtual void __cdecl chrome_pdf::PDFiumEngine::AppendBlankPages(int)" (?AppendBlankPages@PDFiumEngine@chrome_pdf@@UEAAXH@Z)
pdf.dll : fatal error LNK1120: 2 unresolved externals
```

hokein:
@zcbenz 
It looks strange to me. Sorry, I will take a look.

hokein:
Right now I can confirm that the break build is caused by Chrome 43's upgrade.
In Chrome42 , both component and non-component build are fine. Need some time to investigate reason.

PR #117, opened 2015-05-17
Allow file input tag return native file path

hokein:
Fixes atom/electron#1683.

### DIFF
--- a/patches/accelerator.patch
+++ b/patches/accelerator.patch
@@ -1,0 +1,32 @@
+diff --git a/ui/base/accelerators/accelerator.cc b/ui/base/accelerators/accelerator.cc
+index 4bf3b37..d7d075f 100644
+--- a/ui/base/accelerators/accelerator.cc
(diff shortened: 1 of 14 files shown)


paulcbetts:
This is probably a breaking change, we probably need to somehow have developers opt-into this feature, since there is probably a ton of code that searches for `C:\Fakepath` that will now not find it

hokein:
Thanks for the pointing. But the file `<input>` tag's behavior in nw.js also returns the native file path.

zcbenz:
Unlike nw.js which relies on `input` tag for file dialogs, in Electron users should use the `dialog` module to open file dialogs, the `input` tag is only implemented to be compatible with traditional web pages. And as @paulcbetts said, this patch will break many web pages, so I'm closing this.

pietrop:
@zcbenz @paulcbetts What is the future of this feature, is electron ever going to support getting a file path from HTML5` input` element (same as NWJS) ?

PR #119, opened 2015-05-18
ftp: including generated network resource

deepak1556:
[IDR_DIR_HEADER_HTML](https://code.google.com/p/chromium/codesearch#chromium/src/out/Debug/gen/net/grit/net_resources.h&q=dir_header.html&sq=package:chromium&type=cs&l=5) is required for directory listing.


deepak1556:
the required binary is actually shipped.

deepak1556:
This ships `net/grit/net_resources.h`

zcbenz:
Thanks!

PR #120, opened 2015-05-20
patch to provide header field with ResourceRequestDetails

deepak1556:
Is there any style to be followed in creating a patch file, like should the `From:` metadata be removed ? thanks!


hokein:
You can use `git diff <file_name>` to generate the patch.

deepak1556:
Fixed , thanks @hokein

deepak1556:
@zcbenz have made the changes to provide response info.

zcbenz:
Patch launched at https://github.com/atom/libchromiumcontent/commit/abfe09c8758d702dae60f3baa490951d55d89d2b, thanks!

PR #122, opened 2015-05-23
Fix pdf.dll link error in component builds after upgrade to Chrome 43.

hokein:
The link error in [#115](https://github.com/atom/libchromiumcontent/pull/115#issuecomment-104879030)  is caused by https://codereview.chromium.org/980053002/. We need to add printing component dependency in `pdf` library.

Having test on component-build/non-component-build, as well as test printing feature in `electron`.


zcbenz:
Awesome, thanks!

Can you also submit this change to upstream Chromium? It is kind of a bug.

PR #124, opened 2015-05-28
fix creation of httpresponseheaders from raw_headers

deepak1556:
`httpresponseheaders.parse` needs [null](https://code.google.com/p/chromium/codesearch#chromium/src/net/http/http_response_headers.h&sq=package:chromium&l=70&q=http_response_headers) terminated string, this patch fixes it.

### DIFF
--- a/patches/resource_request_details.patch
+++ b/patches/resource_request_details.patch
@@ -1,23 +1,36 @@
 diff --git a/content/public/browser/resource_request_details.cc b/content/public/browser/resource_request_details.cc
-index ad5f47e..0873a32 100644
+index ad5f47e..4e87c72 100644
 --- a/content/public/browser/resource_request_details.cc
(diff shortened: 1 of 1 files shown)


zcbenz:
From Chromium's usages of `raw_headers` I don't think we need this change?
https://code.google.com/p/chromium/codesearch#chromium/src/content/public/common/resource_response.cc&sq=package:chromium&l=17&rcl=1432756788
https://code.google.com/p/chromium/codesearch#chromium/src/content/browser/loader/resource_dispatcher_host_impl.cc&sq=package:chromium&l=754&rcl=1432756788
https://code.google.com/p/chromium/codesearch#chromium/src/content/public/common/resource_response.cc&sq=package:chromium&l=17&rcl=1432756788

deepak1556:
yeah its not required, sorry didnt look into the constructor code previously.. It handles all cases.

PR #125, opened 2015-05-28
Use relative paths in filenames.gypi

aroben:
gyp will convert relative paths to absolute paths for us if we store them in appropriately-named variables. Using gyp's automatic path conversion seems to work better when generating Xcode projects. (For example, gyp was complaining about files from libchromiumcontent being included twice in the project, but that error goes away with this change.)

See https://chromium.googlesource.com/external/gyp/+/master/docs/InputFormatReference.md#Pathname-Relativization for the full explanation of gyp's behavior. The essence is that by naming these variables with `*_files` we get the automatic path conversion behavior.

Also, now that `filenames.gypi` doesn't depend on where libchromiumcontent is extracted, we can generate it at build time and include it in the zip file.

/cc @zcbenz

### DIFF
--- a/script/create-dist
+++ b/script/create-dist
@@ -169,6 +169,7 @@ def main():
       copy_generated_sources(target_arch, component, output_dir)
 
   copy_sources()
+  generate_filenames_gypi()
   create_zip()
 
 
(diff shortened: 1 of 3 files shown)


zcbenz:
Awesome, thanks!

zcbenz:
@aroben I have tried with this, and it turns out that gyp actually converts variables with `*_files` suffix to new relative paths that being relative to current `.gyp` file, instead of converting them to absolute paths. This is also stated in the documents:

> When merging such relative pathnames, they are adjusted so that they can remain valid relative pathnames, despite being relative to a new home.

This is not a problem when building with Xcode, but for ninja builds, using relative paths in `libraries` will cause linking errors because ninja is looking for files under `vendor/brightray/vendor/download/libchromiumcontent/`, however it should look for `../../vendor/brightray/vendor/download/libchromiumcontent/` since it is working under `out/Release`.

I'm not sure if is a bug of gyp or ninja, but this behavior has been there for years. Previously I used to prepend the absolute path of current gyp's dir name to work around it like this: 

```
'libraries': [
  '<(brightray_source_root)/<(libchromiumcontent_library_dir)/libchromiumcontent.so',
],
```

But now since we are using `'<@(libchromiumcontent_shared_libraries)'` to references those libraries, this work-around doesn't work anymore.

I don't know the state of Xcode support in gyp now, but from [Chromium's documents](https://code.google.com/p/chromium/wiki/MacBuildInstructions#Using_Xcode-Ninja_Hybrid) Xcode is not supported for building anymore:

> While using Xcode is unsupported, GYP supports a hybrid approach of using ninja for building, but Xcode for editing and driving compliation. Xcode can still be slow, but it runs fairly well even with indexing enabled.

Maybe we should just switch to use ninja? Or find out a way that makes Xcode work with absolute paths?

I'll revert this for now since it is blocking Electron.

PR #132, opened 2015-06-03
Use the Debug config for shared_library builds

aroben:
Instead of building both shared_library and static_library in Release, we now build shared_library in Debug. This will let applications that link against the shared_library build take advantage of Chromium's assertions that check for things like accessing objects from the proper thread.

This should help with bug reports like https://github.com/atom/electron/issues/951.

/cc @zcbenz

### DIFF
--- a/script/build
+++ b/script/build
@@ -11,7 +11,10 @@ from lib.util import get_configuration, get_output_dir
 SOURCE_ROOT = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
 VENDOR_DIR = os.path.join(SOURCE_ROOT, 'vendor')
 TARGETS = ['chromiumcontent_all']
-COMPONENTS = ['static_library', 'shared_library']
+FLAVORS = [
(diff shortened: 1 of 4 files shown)


zcbenz:
Can you make the Debug config as a new config instead of replacing Release version of shared_library? Currently we are using shared_library Release builds as a way of quick build, like how Chromium makes use of component build, it is used for daily development, and it is required for running on Travis CI since downloading a huge package and extracting it hardly succeeds on Travis CI. And for contributors who don't have a powerful machine or fast internet it is the only way they can build Electron.

We can probably have three kind of binaries: `shared_library`, `static_libary` and `debug_static_library`, and users can choose to download the latter two optionally.

Another technical unrelated reason is, uploading to S3 is very slow outside U.S., since the Windows CI is broken for now, uploading both 32bit and 64bit Windows binaries is already a PITA for me, if we also include the Debug binaries, it is probably impossible for me to upload them.

aroben:
I'm not going to be working on this in the foreseeable future. Please feel free to take this and run with it!

PR #138, opened 2015-06-18
WIP: ARM build

zcbenz:
This is the initial work of ARM build of libchromiumcontent, it still fails for now and I don't have enough time to make it happen. Anyone who is interested in ARM build can start from here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,3 +50,15 @@ If you change `VERSION` to point to a different Chromium release, or modify
     $ script/update
 
 This will regenerate all the project files. Then you can build again.
+
+### Building for ARM target
+
+```bash
+$ ./script/bootstrap
+$ ./script/update -t arm
+$ cd vendor/chromium/src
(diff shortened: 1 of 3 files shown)


deepak1556:
From [common.gypi](https://code.google.com/p/chromium/codesearch#chromium/src/build/common.gypi&q=common.gyp&sq=package:chromium&l=2324) setting `arm_version` sets the required compiler flags ?

zcbenz:
> From common.gypi setting arm_version sets the required compiler flags ?

It should but it doesn't.

zcbenz:
It seems that `standalone_static_library` is not compatible with cross-compilation, to build static_library version of ARM target we have to remove the use of `standalone_static_library` and change how we create distributions.

zcbenz:
Actually `standalone_static_library` is compatible with cross-compilation, it is just that the static libraries generated for host toolset and target toolset have overwritten each other. I think I have found the right solution.

PR #143, opened 2015-06-30
Cross compile for ia32 target

zcbenz:
This PR uses sysroot for building and cross compiles for ia32 target, making it easier to build arm and ia32 targets.


zcbenz:
It seems extremely hard to make the sysroot way work, giving up now.

PR #156, opened 2015-10-02
Ship dependent headers and libraries for desktop capture API.

hokein:
**WIP**


chromiumcontent/chromiumcontent.gyp
```diff
@@ -39,6 +39,8 @@
         '<(DEPTH)/ppapi/ppapi_internal.gyp:ppapi_proxy',
         '<(DEPTH)/ppapi/ppapi_internal.gyp:ppapi_ipc',
         '<(DEPTH)/ppapi/ppapi_internal.gyp:ppapi_shared',
+        '<(DEPTH)/third_party/libyuv/libyuv.gyp:libyuv',
```
zcbenz: `libyuv` is already depended by many components, I don't think we need to add it explicitly.

chromiumcontent/chromiumcontent.gypi
```diff
@@ -122,7 +122,7 @@
           },
         },
       }],
-      ['_target_name in ["gtk2ui", "devtools_http_handler", "devtools_discovery", "http_server"]', {
+      ['_target_name in ["gtk2ui", "devtools_http_handler", "devtools_discovery", "http_server", "desktop_capture", "webrtc_common", "system_wrapper", "rtc_base", "rtc_base_approved", "libyuv"]', {
```
zcbenz: I think only `gtk2ui` should be listed here, it should be a mistake that `devtools_http_handler` and others were included.

script/create-dist
```diff
@@ -59,6 +59,12 @@ BINARIES = {
     'libdevtools_discovery.a',
     'libdevtools_http_handler.a',
     'libhttp_server.a',
+    'libdesktop_capture.a',
```
zcbenz: Can you make the lists in `darwin` and `linux` have the same order? And should `libdesktop_capture_differ_sse2` also be listed here?

script/create-dist
```diff
@@ -68,6 +74,13 @@ BINARIES = {
     'libdevtools_discovery.a',
     'libdevtools_http_handler.a',
     'libhttp_server.a',
+    'libyuv.a',
```
zcbenz: `libyuv.a` doesn't seem to be needed by `desktop_capture`?

chromiumcontent/chromiumcontent.gypi
hokein: We do need it here to guarantee these static libraries are built as stand-alone libraries, otherwise most of the static libraries in Chromium are created as a `thin archive`(at least on Linux) to reduce the size, which will cause the undefined symbol error during `Electron's` linking.

chromiumcontent/chromiumcontent.gyp
hokein: Done.

script/create-dist
hokein: The `libdesktop_capture_differ_sse2` are needed in `webrtc` on Windows and Linux.

chromiumcontent/chromiumcontent.gypi
zcbenz: It was done by [chromiumcontent.gypi#L93-L95](https://github.com/atom/libchromiumcontent/blob/master/chromiumcontent/chromiumcontent.gypi#L93-L95), this rule is to force gtk2ui to build as static library.

hokein: Oh, it's weird. I have meet the linking error if I don't add these libraries at here. I will check it again.

hokein: After a deep look, the code in [chromiumcontent.gypi#L93-L95](https://github.com/atom/libchromiumcontent/blob/master/chromiumcontent/chromiumcontent.gypi#L93-L95) will not be triggered in Debug build(with `component=shared_library`). I think it's fine to remove the `component='static_library'` check for `Debug`/`Release` build there.

zcbenz: :cool:.

hokein: The PR is updated now.

zcbenz:
Looks good to me, currently we are in the process of upgrading the CI machines, so I'm holding on this for a few days until the upgrade is done.

nelix:
Hows this going? I would love to start playing with the capture api.

PR #160, opened 2015-11-08
Add option to build local libchromiumcontent

tejaspathak:
- Currently libchromiumcontent is downloaded by default.
- Now developer can choose to provide local libchromiumcontent src, shared and static path

Tries to solve similar issue noted in: [https://github.com/atom/electron/issues/3310]


zcbenz:
Looks good to me, thanks!

PR #166, opened 2015-12-09
Merge atom branch to master branch

zcbenz:
Now apparently no one else is using libchromiumcontent or brightray except for Electron, maintaining an extra atom branch is like hell and let's just merge it to master branch.


deepak1556:
:+1:

PR #175, opened 2016-01-26
Upgrade to Chrome 49

zcbenz:
* [x] Update all the patches
* [x] Compile on Linux
* [x] Compile on OS X
* [x] Compile on Windows
* [x] Generate distributions on Linux
* [x] Generate distributions on OS X
* [x] Generate distributions on Windows
* [x] Check everything works with Electron


zcbenz:
We are meeting a :hankey: problem, Node will not release new version for V8 4.8 (and 4.9 in future):

```irc
<zcbenz>	Hi, will Node v6 be released for V8 4.8 update?
<Trott>		 Node v6 will be released in April.
<Trott>		 See https://github.com/nodejs/LTS for major version release schedule.
<Trott>		(Especially the timeline all the way at the bottom.)
<zcbenz>	Trott: thanks, I saw that V8 4.8 update will not be merged into Node v5, so does it mean that we will not get Node releases with new V8 versions until 3 months later, when Node v6 gets released?
<evanlucas>	zcbenz that is the idea...generally, a v8 minor bump will result in a node major bump
<zcbenz>	evanlucas: there is problem with Node's schedule, v8 bumps minor version about every 6 weeks, while Node bumps major version every 4 months, so Node will skip 2 v8 versions between major releases.
<evanlucas>	correct...the relationship between node major versions and v8 minor versions will not be 1-to-1
<zcbenz>	cool, understood
```

so it would mean we are going to use the master branch of Node for 3 months while taking care of V8 bugs ourselves.

/cc @kevinsawicki @jlord

bpasero:
Ouch :-|

> v8 minor bump will result in a node major bump
if that is true then I wonder why node v5 went from V8 45 to V8 47.

kevinsawicki:
@zcbenz so this will be something that will happen often then right? We'll have to use node master whenever we are upgrading chrome and are in between node major releases?

Has electron used node master in the past for long periods of time and was it a lot to maintain?

zcbenz:
> @zcbenz so this will be something that will happen often then right? We'll have to use node master whenever we are upgrading chrome and are in between node major releases?

Yeah.

> Has electron used node master in the past for long periods of time and was it a lot to maintain?

We used to use Node 0.11 and io.js pre-1.0 for a very long time, that was painful. It is not hard to maintain now because Node updates V8 version in their master branch.

zcbenz:
To people watching this pull request: I'm gonna take a vacation, so please do not expect this to be done until the end of February.

maxogden:
Have fun and get some rest!

bpasero:
:+1: enjoy. I am actually also on vacation ;)

ijuma:
Would skipping Chrome releases be an option?

paulcbetts:
> Would skipping Chrome releases be an option?

Sure, but the work to move to Chromium 49 is almost certainly going to incrementally build off of the work to move to Chromium 48

diorahman:
Have fun @zcbenz!

bbondy:
Is this still coming end of February? Thanks for the update.

zcbenz:
I'm unlikely to make it this month, recently I'm very busy moving to a new country, so I'll probably skip Chrome 48 and do Chrome 49 instead.

bbondy:
Could you overview what is remaining in case someone else can pick it up?
Or is it only per the above message:
> Check everything works with Electron 

In which case we might just pull this into our fork only for now.

YuriSolovyov:
I know this kind of the same as node master, but what about node nightlies?

jsnar:
Chromium 49 was released.... and Electron is still on Chromium 47. In the meanwhile nw.js was updated to Chromium 49 from 48... 

Is there any update here? Brave browser really need this to fix security issues.

lipis:
:100:

zcbenz:
This branch has been updated with Chrome 49.

paulcbetts:
@zcbenz Set up a dummy PR with Brightray and Atom pointing to this build and I'll help work through the Electron changes!

zcbenz:
@paulcbetts Feel free to add commits to https://github.com/atom/electron/pull/4714.

paulcbetts:
@zcbenz Alright, made some progress in https://github.com/atom/electron/pull/4716 and https://github.com/atom/brightray/pull/201

bpasero:
@zcbenz is the idea to stay on node.js 5.1.1 even though Chrome was updated?

zcbenz:
@bpasero I'll update Node to latest stable in near future, it is not updated because it is working well and there is no strong reason to update it.

bpasero:
:+1:

jsnar:
I'm surprised to see "no strong reason" to update Node.js. At least there is openssl update in the new Node.js versions, which fixed a bunch of security issues: https://github.com/nodejs/node/blob/v5.8.0/CHANGELOG.md#notable-changes-1

Issue #178, opened 2016-02-15
libchromiumcontent/Electron is licensed incorrectly, current build is GPL


MarshallOfSound:
The current build process of electron builds a custom version of chromium through the module [libchromiumcontent][0].  This is all very well but there is a single flag inside this build process found [here][1] that turns your build of chromium into GPL licensed code.  

That flag builds ffmpeg with a whole load of additional codecs which are licensed under GPL.  An example of such a codec is `H.264`.  It has been confirmed that `libchromiumcontent` builds this codec in [this issue][2] and confirmed that it is GPL licensed on [ffmpeg's site][3].  This codec is just one example of a GPL component included in the `ffpmeg` build `libchromiumcontent` generates.  

### Problem

Electron is currently licensed under the [MIT License][4] which is against the terms of GPL.  In lamens terms, anything that uses GPL licensed software must also be GPL licensed.  This is a MASSIVE issue for many companies that use electron to ship commercial grade applications.  It means that any application they ship can be claimed to have to be GPL and they MUST release their source code.

Further evidence of the GPL nature of this build style is found in the `nw.js` documentation [here][5].

### Solution

We would like to request that you ship `non-GPL` versions of the prebuilt electron packages and update your licenses / README to reflect the fact that some packages use GPL licensed software.  

We have attached a PR [here][6] (#179) to help fix this issue.  This PR basically moves the change in [this commit][1] to an environment variable so you can easily turn on / off the GPL version.

[0]: https://github.com/atom/libchromiumcontent
[1]: https://github.com/atom/libchromiumcontent/pull/69/files#diff-2544685f4860e4633fd5747ed88ec2caR6
[2]: https://github.com/atom/libchromiumcontent/issues/174#issuecomment-175297780
[3]: https://www.ffmpeg.org/doxygen/2.6/md_LICENSE.html
[4]: https://github.com/atom/electron/blob/master/LICENSE
[5]: http://docs.nwjs.io/en/v0.13.0-beta5/For%20Developers/Enable%20Proprietary%20Codecs/#enable-proprietary-codecs-in-nwjs
[6]: https://github.com/atom/libchromiumcontent/pull/179

paulcbetts:
The comments in https://github.com/atom/libchromiumcontent/pull/69 seem to explicitly indicate that this isn't true (despite the flag being the same name)

MarshallOfSound:
[...]

PR #179, opened 2016-02-15
Move GPL causing flags to en environment variable check

MarshallOfSound:
The flag `proprietary_codecs` causes the resulting build of `ffmpeg` to be GPL licensed.  Which in turn makes `chromium` GPL licensed and will then in turn make `electron` GPL.

This has been outlined in details here #178

[Review on Reviewable](https://reviewable.io/reviews/atom/libchromiumcontent/179)


zcbenz:
Closing because of https://github.com/atom/libchromiumcontent/pull/69/files#issuecomment-184186331.

PR #180, opened 2016-02-18
Prevent capabilities from being dropped in renderer threads in Linux

jviotti:
Chromium automatically drops all capabilities of renderer threads in
Linux, which may cause issues in a context like Electron, where the main
and renderer threads are supposed to keep inherited permissions over the
system.

See https://github.com/atom/electron/issues/3666


zcbenz:
:+1:

PR #183, opened 2016-02-26
fix OS X 10.11 build

bridiver:
these diffs were pulled from various chromium bugs and fix the build on 10.11. I don't have a machine available with 10.10 to make sure it is backwards compatible


zcbenz:
Currently building Chromium requires OS X 10.10 SDK, if you are using Xcode 7 then it is very likely you are using 10.11 SDK.

I'm opposed to adding patch to just fix building for 10.11 SDK, maintaining patches requires a lot of work when upgrading Chrome version.

bridiver:
As I mentioned, these patches came from Chromium and this just backports them to 47 so there is nothing to maintain

zcbenz:
Sorry I didn't notice that, backporting patches is good to me.

PR #184, opened 2016-03-05
Add patch to disable suspending animation

zcbenz:
Refs https://github.com/atom/electron/issues/4317#issuecomment-192619978.

(I'm not sure whether it would work.)


zcbenz:
Not working. :(

PR #187, opened 2016-03-25
Update web_contents_impl.patch

deepak1556:
Minimal patch to obtain disposition with `ShouldCreateWebContents`.

### DIFF
--- a/patches/web_contents_impl.patch
+++ b/patches/web_contents_impl.patch
@@ -1,8 +1,24 @@
 diff --git a/content/browser/web_contents/web_contents_impl.cc b/content/browser/web_contents/web_contents_impl.cc
-index b5b0b78..aac6876 100644
+index b5b0b78..1c42b6c 100644
 --- a/content/browser/web_contents/web_contents_impl.cc
 +++ b/content/browser/web_contents/web_contents_impl.cc
-@@ -4353,6 +4353,7 @@ void WebContentsImpl::RendererUnresponsive(
+@@ -1820,10 +1820,14 @@ void WebContentsImpl::CreateNewWindow(
+       static_cast<SessionStorageNamespaceImpl*>(session_storage_namespace);
+   CHECK(session_storage_namespace_impl->IsFromContext(dom_storage_context));
+ 
++  WindowContainerType window_container_type = WINDOW_CONTAINER_TYPE_NORMAL;
++  if (params.disposition == NEW_BACKGROUND_TAB)
++    window_container_type = WINDOW_CONTAINER_TYPE_BACKGROUND;
++
+   if (delegate_ &&
+       !delegate_->ShouldCreateWebContents(
+           this, route_id, main_frame_route_id, main_frame_widget_route_id,
+-          params.window_container_type, params.frame_name, params.target_url,
++          window_container_type, params.frame_name, params.target_url,
+           partition_id, session_storage_namespace)) {
+     if (route_id != MSG_ROUTING_NONE &&
+         !RenderViewHost::FromID(render_process_id, route_id)) {
+@@ -4353,6 +4357,7 @@ void WebContentsImpl::RendererUnresponsive(
  
    if (rfhi->is_waiting_for_beforeunload_ack() ||
        rfhi->IsWaitingForUnloadACK()) {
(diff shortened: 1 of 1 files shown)


patches/web_contents_impl.patch
```diff
@@ -1,8 +1,24 @@
...
+index b5b0b78..1c42b6c 100644
 --- a/content/browser/web_contents/web_contents_impl.cc
 +++ b/content/browser/web_contents/web_contents_impl.cc
-@@ -4353,6 +4353,7 @@ void WebContentsImpl::RendererUnresponsive(
+@@ -1820,10 +1820,14 @@ void WebContentsImpl::CreateNewWindow(
+       static_cast<SessionStorageNamespaceImpl*>(session_storage_namespace);
+   CHECK(session_storage_namespace_impl->IsFromContext(dom_storage_context));
+ 
++  WindowContainerType window_container_type = WINDOW_CONTAINER_TYPE_NORMAL;
++  if (params.disposition == NEW_BACKGROUND_TAB)
```
bridiver: I don't believe this is correct. A background tab is a tab that opens without becoming the active tab. A background window is a background extension process. The disposition is available in AddNewContents and CanCreateWindow which we added in our fork. @zcbenz brought up some issues which we have addressed and I will submit another PR when I have more time, but you can look at https://github.com/brave/electron/blob/master/atom/browser/api/atom_api_web_contents.cc to see what we've done.

deepak1556: Yup true, in electron atleast currently we dont rely on the content layer to create the webContents. `AddNewContents` wont get triggered unless you allow the `content_layer` to create and `CanCreateWindow` is useless to associate with the webcontents we are gonna create, as the frame_id are gonna be obselete once we return `false` in `ShouldCreateWebContents`. 

If `ShouldCreateWebContents` provided the navigation policy, this patch woulnt have been required, but if were to add another param to the method I would be changing a few other files which is really unnecessary and will become hard to maintain.

Associating `WindowContainerType` and `disposition` are wrong, but this is the minimal way i could send across the navigation policy to dispatch in `new-window` event.

bridiver: another option is to use CanCreateWindow (in AtomBrowserClient) instead of ShouldCreateWebContents. I have been meaning to take a closer look at that option. I'm pretty sure it will "just work", but I haven't had time to take a closer look. You can retrieve the content webcontents using RenderFrameHost::FromID and then get the api wrapper for the emit call.

deepak1556: Damn, last time overlooked the `render_process_id`. Should work, let me verify and close this patch. Thanks!

deepak1556: `frame_name` not available with `CanCreateWindow`, its like missing one parameter for the other :/ .

> Patching web_contents_impl to change the meaning of the window container seems like a future backwards compatibility issue to me.

Yes even i hate to change the meaning, but i approached the upstream mailing-list before creating this patch https://groups.google.com/a/chromium.org/forum/#!topic/chromium-dev/J3obbDcrbhA , couldnt get any feedback. As far as i can see we dont rely on `window_container_type` from `shouldcreatewebcontents`. If in future we decide to delgate to `content_layer` then this patch can be removed.

bridiver: is it available from the RenderFrameHost at that point? `GetFrameName`?

bridiver: The strange thing is that Chrome uses that method for popup blocking and as far as I can tell it is lost in the process. Maybe this is actually a chrome bug and the frame_name should be added to the method?

bridiver: tested on Chrome and frame_name is lost when opening a blocked popup, but is not lost if there is an exception in the preferences to allow the popup.

deepak1556: >  frame_name should be added to the method?

Yup, provide `frame_name` with `CanCreateWindow` or add `disposition` to `ShouldCreateWebContents`. @zcbenz which method would you prefer to be patched ?

bridiver: `frame_name` to CanCreateWindow looks like something that can likely be upstreamed to Chrome since it appears to be a legitimate bug affecting popup handling. In Chrome it should also be added to `BlockedWindowParams`

bridiver: @deepak1556 if we can work together on this one I think I can resolve the remaining issues between our fork and the main repo. We don't want to diverge from what you are doing and we have extensions mostly working which will close #98

deepak1556: Sure, would patching `CanCreateWindow` help ? the link you provided points to something other issue.

bridiver: Yea, I think patching `CanCreateWindow` will give us everything we need. We completely removed node from  renderers with node disabled, sandboxed them and added scriptable opener support and content scripts for them. We also now have extensions working with background pages. If we can capture the opener info in `CanCreateWindow` we can create a completely backwards compatible version that doesn't cause problems for node-enabled renderers that have to be restarted on every navigation right now

deepak1556:
closing in favor of https://github.com/atom/libchromiumcontent/pull/191

PR #188, opened 2016-03-26
replace linux_use_tcmalloc with use_allocator variable

deepak1556:


zcbenz:
Maybe we should just remove the flag? It seems that the building works fine without it.

deepak1556:
Yup thought about that, i would like to build electron with sanitizers but tcmalloc and sanitizers dont work well. I can work with local modified version when needed, so your call.

zcbenz:
Let's use tcmalloc by default then, I'll remove the flag in another commit.

PR #194, opened 2016-04-05
Add patch to prevent render widget from getting hide

zcbenz:
Refs https://github.com/electron/electron/issues/4925.


deepak1556:
:+1:

PR #195, opened 2016-04-07
add some media strings to content strings resources

deepak1556:
content layer doesnt ship audio strings which media resources depend on, chrome ships it in `generated_resources.grd`.


deepak1556:
Wasn't sure if its required to provide localized string for default audio device labels. Just following chromes' behavior.

deepak1556:
Not needed, since we are not localizing the default audio label strings.

PR #196, opened 2016-04-12
ship security_state component

deepak1556:


patches/content_strings.patch
```diff
@@ -0,0 +1,43 @@
...
+index e56c685..4eb9ba2 100644
+--- a/content/app/strings/content_strings.grd
++++ b/content/app/strings/content_strings.grd
+@@ -609,6 +609,38 @@ below:
+         Please enter a number.
+       </message>
+ 
++      <!-- Strings describing Chrome security policy for DevTools security panel -->
++      <message name="IDS_MAJOR_SHA1" desc="Summary phrase for a security problem where the site's certificate expires in 2017 or later and contains a SHA1 signature in the chain." translateable="false">
++        SHA-1 Certificate
```
zcbenz: We can probably just return raw string in `GetSecurityStyle`.

script/create-dist
```diff
@@ -106,6 +108,7 @@ BINARIES = {
     os.path.join('obj', 'components', 'devtools_http_handler.lib'),
     os.path.join('obj', 'components', 'cdm_renderer.cc.pdb'),
     os.path.join('obj', 'components', 'cdm_renderer.lib'),
+    os.path.join('obj', 'components', 'security_state.lib'),
```
zcbenz: `security_state.cc.pdb` should also be added.

deepak1556:
Have made the changes, thanks!

zcbenz:
👍

PR #201, opened 2016-05-05
add patch to make url::AddStandardScheme thread safe

deepak1556:


patches/url_util.patch
```diff
@@ -0,0 +1,75 @@
...
+   InitStandardSchemes();
+-  for (size_t i = 0; i < standard_schemes->size(); i++) {
+-    if (base::LowerCaseEqualsASCII(
+-            typename CharToStringPiece<CHAR>::Piece(
+-                &spec[scheme.begin], scheme.len),
+-            standard_schemes->at(i).scheme)) {
+-      *type = standard_schemes->at(i).type;
+-      return true;
++  {
++    base::AutoLock auto_lock(lock);
```
zcbenz: Can you keep the original indention so we can make the patch minimal?

```c++
{
base::AutoLock auto_lock(lock);
for (size_t i = 0; i < standard_schemes->size(); i++) {
  ...
}
}
```

deepak1556: Done, thanks!

zcbenz:
👍

zcbenz:
@deepak1556 This is probably going to cause big performance issues, since every operation of URL parsing would involve acquiring the lock.

deepak1556:
I am not able to provide a test case for which this patch would be necessary, we just need to make sure users register standard schemes before they do any content load in the app. It should be safe to call `AddStandardScheme`. We can revert this patch, sorry for not testing it properly.

zcbenz:
@deepak1556 I think it is because we are using the Release version of libchromiumcontent, and the  check for `standard_schemes_locked` gets invalidated. We should require users to run `protocol.registerStandardSchemes` before the `ready` event of `app` then, otherwise race conditions may happen after the networking thread gets started.

deepak1556:
Currently `protocol` module is not allowed to be used before `ready` event, should we change this or instead we can make this an api of `app`(Shouldn't break any existing applications).?

zcbenz:
Yeah we need to make `protocol` module available before the `ready` event, while prevent its APIs other than `registerStandardSchemes` from being called at that time.

PR #206, opened 2016-05-09
Include mojo in component build

paulcbetts:
This is needed for https://github.com/electron/electron/pull/5413 to build in Debug mode, thanks to @deepak1556 who actually did this PR, I'm just the guy applying the patchfile :)


deepak1556:
After chrome50
```
<(DEPTH)/third_party/mojo/mojo_public.gyp => <(DEPTH)/mojo/mojo_public.gyp
```

I think we should wait for the upgrade, also the static libraries need to be shipped and linked in brightray.

paulcbetts:
> I think we should wait for the upgrade

Agree, I'll retarget this to chrome50

paulcbetts:
Don't need it!

PR #226, opened 2016-08-02
Offscreen rendering support for Electron

gerhardberger:
This PR adds the necessary includes and patches to Chromium that are needed for offscreen rendering in Electron.

The [PR](https://github.com/electron/electron/pull/6691) in Electron.


zcbenz:
👍 

I did some modifications on `web_contents.patch` to make the diff minimal.

PR #227, opened 2016-08-30
Fix for webviews using default (empty) BeginFrameSubscription implementation

brenca:
This PR adds a small patch to fix electron/electron#5665
`RenderWidgetHostViewChildFrame` was missing some functions, I tried to keep the implementation as small as possible for easier maintenance.

### DIFF
--- a/patches/guest_view.patch
+++ b/patches/guest_view.patch
@@ -1,0 +1,67 @@
+diff --git a/content/browser/frame_host/render_widget_host_view_child_frame.cc b/content/browser/frame_host/render_widget_host_view_child_frame.cc
+index 7101a90..8349471 100644
+--- a/content/browser/frame_host/render_widget_host_view_child_frame.cc
++++ b/content/browser/frame_host/render_widget_host_view_child_frame.cc
+@@ -28,8 +28,10 @@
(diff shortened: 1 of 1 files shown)


IsaiahJTurner:
@brenca I think something happened with copy and pasting source changes.

brenca:
@IsaiahJTurner What do you mean?

IsaiahJTurner:
Take that back. I hadn't seen a `.patch` before. I'll test it shortly, thanks for doing this!

brenca:
:) It was actually an issue with chromium's code, this is why I made a patch file to fix it. Also, you might want to wait until @zcbenz pulls this in (if he will :) ) and it gets included in the next electron release, since building a custom `libchromiumcontent` can easily take a full day (and most of it is just the building!), if you don't have machine specifically made for this kind of stuff. I couldn't test this out on Mac, but I did test it on Windows (the screenshot I added) and the subscription seems to be working perfectly now.

zeke:
💰

gerhardberger:
I tried it out on macOS and works like a charm! 👌

zcbenz:
👍

PR #236, opened 2016-10-09
update render_message_filter.patch

deepak1556:
This exposes `ResourceRequestBody` in `BrowserClient::CanCreateWindow` , so that we can initiate POST requests for browser created windows.


zcbenz:
👍

deepak1556:
@zcbenz seems like build wasn't generated for platforms other than linux ?

zcbenz:
I have triggered a rebuild, hopefully it will finish within a day.

PR #239, opened 2016-11-02
Upgrade to Chrome 54

zcbenz:
- [x] Switch to gn
- [x] Update all the patches
- [x] Compile on Linux
- [x] Compile on OS X
- [x] Compile on Windows
- [x] Generate distributions on Linux
- [x] Generate distributions on OS X
- [x] Generate distributions on Windows
- [x] Check how to make widevine compile
- [x] Check everything works with Electron


kevinsawicki:
Hey @zcbenz, there was an issue on our Linux CI where the logs were too long because of an assertion that was getting logged thousands of times. I fixed it on master in https://github.com/electron/libchromiumcontent/pull/246 so you may want to rebase/merge that into here to get the `libchromiumcontent-linux` build working.

script/update
```diff
@@ -237,7 +237,7 @@ def gn_args(target_arch, component, gyp_defines):
   if component == 'shared_library':
     args.append('is_component_build=true')
   else:
-    args.append('is_component_build=false is_official_build=true')
+    args.append('is_component_build=false')
```
poiru: This seems like it could potentially degrade performance. Out of curiosity, how was the libv8_base library corrupted?

zcbenz: It would generate `LNK1107` error when generating `libv8_base.lib`, by removing the `/GL` linker parameter which is added by `is_official_build=true`, the error is gone.

Chromium does not have this error because v8 is a `source_set` in it.

demian85:
Is it safe to build using this branch? #252

kevinsawicki:
> Is it safe to build using this branch?

This branch will (most likely) be merged and released this month in Electron 1.5.0, but there will be releases before that done from master and released as Electron 1.4.x.

So if it is easier to do what you are trying to do against Chrome 54 and that timeline works for you then building against this branch is fine.

zcbenz:
Closed in favor of https://github.com/electron/libchromiumcontent/pull/254.

PR #243, opened 2016-11-08
Upgrade to Chrome 53.0.2785.143

kevinsawicki:
Similar to #242 but cherry-picked on top of the commit that electron is currently built against, 82751b1.

This ensures that the upgrade in electron/electron will only be this chrome upgrade and not the other changes landed to libchromiumcontent's master already that may not be ready for release.

This pull request will not be merged directly but the SHA-1 of the commit will be used to bump the submodule in brightray.

Refs https://github.com/electron/brightray/pull/259


kevinsawicki:
Closing this out, was able to apply master here to master on electron pretty easily in https://github.com/electron/electron/pull/7906

PR #249, opened 2017-01-03
Use GN instead of GYP to build Chromium sources

alespergl:
This is not yet a full migration to GN. It still doesn't cover all platforms and configurations. More changes need to be done to make it a full replacement of the old GYP based build, but I want to let others to have a look so that we can have a discussion about the approach.

**A few notes about this approach** (and how it differs from what #239 does):

- Main point: no changes to the GN tool

- It's limited to the libchromiumcontent repo, no changes needed in other repos.
  > Note: Electron needs a few small changes that I consider bug fixes. More on that below.

- It's backwards compatible with the current pipeline: build libcc -> create-dist -> run `generate_filenames_gypi` -> build Electron

- I wanted to keep Chromium's GN build files work the same when not configuring for Electron build. It should remain possible to build Chromium normally even with our patches applied.

- I made use of the possibility to specify build arguments in static .gn files (in the `chromiumcontent/args` subdirectory) and import them when running `gn gen`. This allows easier custom configurations (e.g. various Chromium features enabled/disabled, like webrtc or others)

- I used a trick to build static libraries out of the .obj files produced by the `source_set` targets. I let the build run, then re-run it to build the libs (more info in `chromiumcontent/BUILD.gn`). This is handy to keep compatibility with the existing pipeline, but it's not necessary. We could just distribute the .obj files and link them to Electron (tried it and works). That would require a small change to Electron's GYP files, but the end result would be the same.

- The statically linked executables `mksnapshot` and `chromedriver` need to be built in a configuration which doesn't tweak ICU and V8 to make them exportable from `node.dll`. That config can be the "FFMPEG" one and I mention that plan in TODO comments.

**Changes to the Electron repo**

- With this change, [this workaround](https://github.com/electron/electron/blob/master/atom/common/node_bindings.cc#L66-L72) can be removed

- I wanted to make sure that every source file is built with the right set of defines which don't conflict with one another. So in case of V8, it's built with `BUILDING_V8_SHARED` and consumed with `USING_V8_SHARED`. Therefore the latter needs to be added to Electron's GYP file. 

**Stuff that needs to be done still**

- This now works on Windows and Mac with the `static_library` build. There are some Mac linker questions I need help with.
- Linux `static_library` and `shared_library` are not done yet.
- I need to understand the purpose of the `ffmpeg` config.

I fully acknowledge that something about these changes may not work in some important scenario that I am not aware of. Please let me know if that is the case.

### DIFF
--- a/patches/build_gn.patch
+++ b/patches/build_gn.patch
@@ -19,3 +19,15 @@ index e5fdb29e858f..01b6973cb494 100644
        ldflags = [
          # Not to strip important symbols by -Wl,-dead_strip.
          "-Wl,-exported_symbol,_PPP_GetInterface",
+diff --git a/third_party/WebKit/Source/platform/BUILD.gn b/third_party/WebKit/Source/platform/BUILD.gn
+index 631aa54..84422a3 100644
+--- a/third_party/WebKit/Source/platform/BUILD.gn
++++ b/third_party/WebKit/Source/platform/BUILD.gn
+@@ -172,6 +172,7 @@ action("character_data") {
+ }
+ 
+ executable("character_data_generator") {
++  configs += [ "//chromiumcontent:build_time_executable" ]
+   sources = [
+     "text/CharacterPropertyDataGenerator.cpp",
+     "text/CharacterPropertyDataGenerator.h",
--- a/patches/build_gn.patch
+++ b/patches/build_gn.patch
@@ -1,0 +1,21 @@
+diff --git a/third_party/widevine/cdm/BUILD.gn b/third_party/widevine/cdm/BUILD.gn
+index e5fdb29e858f..01b6973cb494 100644
+--- a/third_party/widevine/cdm/BUILD.gn
++++ b/third_party/widevine/cdm/BUILD.gn
+@@ -2,6 +2,7 @@
+ # Use of this source code is governed by a BSD-style license that can be
+ # found in the LICENSE file.
+ 
++import("//chromiumcontent/config.gni")
+ import("//build/config/chrome_build.gni")
+ import("//build/config/features.gni")
+ import("//chrome/version.gni")  # TODO layering violation
+@@ -93,7 +94,7 @@ if (widevine_cdm_binary_files != []) {
+       "//build/config/sanitizers:deps",
+     ]
+ 
+-    if (is_mac) {
++    if (is_mac && !is_electron_build) {
+       ldflags = [
+         # Not to strip important symbols by -Wl,-dead_strip.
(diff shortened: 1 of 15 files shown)


patches/build_gn.patch
```diff
@@ -0,0 +1,21 @@
...
+ import("//chrome/version.gni")  # TODO layering violation
+@@ -93,7 +94,7 @@ if (widevine_cdm_binary_files != []) {
+       "//build/config/sanitizers:deps",
+     ]
+ 
+-    if (is_mac) {
++    if (is_mac && !is_electron_build) {
+       ldflags = [
+         # Not to strip important symbols by -Wl,-dead_strip.
+         "-Wl,-exported_symbol,_PPP_GetInterface",
```
bridiver: for mac you will need `-Wl,-install_name,@loader_path/libwidevinecdm.dylib`
gyp handled it correctly, but there is a comment in the gn file `#TODO(jrummell) Mac: 'DYLIB_INSTALL_NAME_BASE': '@loader_path',`

script/update
```diff
@@ -194,6 +180,43 @@ def run_gyp(target_arch, defines):
...
+  if sys.platform in ['win32', 'cygwin']:
+    platform = 'win32'
+    key = 'win/gn.exe.sha1'
+  elif sys.platform == 'darwin':
+    platform = 'darwin'
+    key = 'mac/gn.sha1'
+  else:
+    platform = 'linux*'
+    key = 'linux64/gn.sha1'
+  subprocess.call([sys.executable, download, '--no_resume',
```
kevinsawicki: Should the result here be returned so the build fails when this command fails?

alespergl:
Above are the commits which make the build work on all platforms and configurations. I also created separate PRs for **brightray** and **electron** to accompany this PR, which are to be applied if this goes through.

zcbenz:
Closed in favor of https://github.com/electron/libchromiumcontent/pull/254.

CapOM:
So there is no static and shared libchromiumcontent zips anymore ? Replaced by SCCACHE ? When building master branch of electron there is no step that download these zips anymore.

nornagon:
Current master of electron doesn't use libchromiumcontent any more, this repo is deprecated. See https://github.com/electron/electron/blob/master/docs/development/build-instructions-gn.md

CapOM:
Thx that was my guess but I found the doc not clear about it, espcially for a dev that was building previous version of electron. This is just my feeling but the build instruction are less clear now that it used to be but hopefuly it is just me :)

nornagon:
Yep, things are a little bit more complicated now than they used to be, but I think it'll get better over the next few months as we build tooling around the new build process!

Issue #253, opened 2017-01-10
Fix for MediaRecorder on Mac


demian85:
Hi,

I reported [this issue](https://bugs.chromium.org/p/chromium/issues/detail?id=678269) 5 days ago and someone [published a patch](https://codereview.chromium.org/2617143003) targeted for Chrome 57.
I've been struggling with this issue for at least 2 weeks. My project development is currently stuck because it depends on this feature to work.

As you can see in the issue report, I attached [a chrome extension demo](https://bugs.chromium.org/p/chromium/issues/attachment?aid=266198) that confirms bug is fixed in latest Chromium build with patch applied.

Please, I would like to know what it takes for this patch to be applied to the next libchromiumcontent release without the need of having to compile from source myself (tried and it's not working... #252)

Thanks.

kevinsawicki:
@demian85 can you open a pull request in this repository that adds the patch to the `patches` folder and we can take a look at the scope of it?

kevinsawicki:
You can see an example of another patch pull request at https://github.com/electron/libchromiumcontent/pull/251

demian85:
Great, thanks!

kevinsawicki:
Fixed by #255

demian85:
Well, this is disturbing. I've just downloaded electron 1.4.15 and bug is still there -_-

kevinsawicki:
> Well, this is disturbing. I've just downloaded electron 1.4.15 and bug is still there -_-

Could there be another patch or change that needed to go with it?

demian85:
Well, I just tested on my compiled development version and it works... But release version does not. Wtf? Can you please explain what's the difference and how can I create a distributable release of electron? I have a bunch of files and I don't know where to put them. Also, my compiled version is based on Chrome 54. Perhaps that has something to do? When is the electron based on Chrome 54 release planed?

demian85:
ping @kevinsawicki 

I keep playing with electron to see if issue still exists or no and where.
I just found out that my compiled version based on chrome 54 works, chrome canary too, but current chrome stable does not work.
So perhaps the issue depends on another fix already in chrome 54, that's why using the latest electron stable build does not work, it's based on chrome 53, right?

What I also see in my console is that every webm blob chunk has mime type assigned, whereas on Chrome 53 it doesn't. Suspicious.

So, could you please tell me when the new version of electron will be released?

Thanks.

kevinsawicki:
[...]

PR #255, opened 2017-01-13
Add patch for WebmMuxer

demian85:
@see #253 
@see  https://codereview.chromium.org/2617143003

Thanks.

### DIFF
--- a/patches/webm_muxer_fix.patch
+++ b/patches/webm_muxer_fix.patch
@@ -1,0 +1,14 @@
+Index: media/muxers/webm_muxer.cc
+diff --git a/media/muxers/webm_muxer.cc b/media/muxers/webm_muxer.cc
+index 9a4819a4ff68ec600f5a70116a694095a9fe02c5..a0890903bc667fc3e619d651e3e1215c284c179f 100644
+--- a/media/muxers/webm_muxer.cc
++++ b/media/muxers/webm_muxer.cc
+@@ -236,8 +236,6 @@ void WebmMuxer::AddVideoTrack(const gfx::Size& frame_size, double frame_rate) {
+   DCHECK_EQ(0ull, video_track->crop_bottom());
+   DCHECK_EQ(0.0f, video_track->frame_rate());
+ 
+-  video_track->set_default_duration(base::Time::kNanosecondsPerSecond /
+-                                    frame_rate);
+   // Segment's timestamps should be in milliseconds, DCHECK it. See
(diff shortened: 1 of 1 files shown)


kevinsawicki:
The patch provided does not appear to apply cleanly to Chrome 53.0.2785.143

The error when running `script/update` on this branch:


```
source file is different - media/muxers/webm_muxer.cc
webm_muxer_fix.patch failed to apply
```

demian85:
How is that possible? I specifically checked out that tag and applied patch with `git apply` and worked fine.
```
MacRamboPro:chromium-53 demian$ git log
commit 12ae4ebf489873fc4aad9efc231b9151b330a938
Author: chrome-cron <chrome-cron@google.com>
Date:   Sun Sep 25 01:36:29 2016 +0000

    Publish DEPS for Chromium 53.0.2785.143
MacRamboPro:chromium-53 demian$ git tag -l
53.0.2785.143
```

What should I do? Is that commit the correct one?

kevinsawicki:
How is that possible? I specifically checked out that tag and applied patch with git apply and worked fine.

I think the patch library uses to apply them in `libchromiumcontent` has stricter SHA checks.

So if you apply the patch to a `53.0.2785.143` checkout using `git apply` and then run `git diff`, you'll see the updated patch.

Here is it for:


```patch
diff --git a/media/muxers/webm_muxer.cc b/media/muxers/webm_muxer.cc
index 0633baf..c134f77 100644
--- a/media/muxers/webm_muxer.cc
+++ b/media/muxers/webm_muxer.cc
@@ -234,8 +234,6 @@ void WebmMuxer::AddVideoTrack(const gfx::Size& frame_size, double frame_rate) {
   DCHECK_EQ(0ull, video_track->crop_bottom());
   DCHECK_EQ(0.0f, video_track->frame_rate());
 
-  video_track->set_default_duration(base::Time::kNanosecondsPerSecond /
-                                    frame_rate);
   // Segment's timestamps should be in milliseconds, DCHECK it. See
   // http://www.webmproject.org/docs/container/#muxer-guidelines
   DCHECK_EQ(1000000ull, segment_.GetSegmentInfo()->timecode_scale());
```

demian85:
So, updated patch. Does it work now?

kevinsawicki:
Thanks for this 👍

kevinsawicki:
@demian85 this was released in Electron 1.4.15

PR #256, opened 2017-01-18
Ship necessary pdf libraries on all platforms

deepak1556:


zcbenz:
I have rebased this branch and updated the patch for Chrome 56.

deepak1556:
Thanks @zcbenz

PR #257, opened 2017-01-19
Enable is_official_build for static_library builds

poiru:
This turns on whole program optimization.


poiru:
>Can you verify all the Windows builds of Electron (x64/ia32 - Debug/Release) work with this flag on? The last time I tried it, the linker crashed on Windows with LNK1107.

Will do!

>And if it works, I think we can also enable it for shared_library builds?

Would there be any benefit from doing this since shared_library builds are supposed to only be used for local development? Improved performance at the expense of link time might not be the right tradeoff for development builds.

poiru:
CI is failing with `fatal error LNK1180: insufficient disk space to complete link` on Windows 😞 Maybe splitting the build multiple Jenkins jobs (by static/shared and/or ia32/x64) would help here?

@kevinsawicki @zcbenz What do you think?

kevinsawicki:
> fatal error LNK1180: insufficient disk space to complete link

How much space if this taking up for you locally on Windows? Even if we split it up, the Jenkins workspaces are on the same machine and reused between builds.

poiru:
@kevinsawicki My build resulted in:

* `out` and `out_32`: 55GB each
* `out_component` and `out_component_32`: 15GB each
* `out_ffmpeg`: 4GB

So a total of ~144GB for the `out*` directories plus ~3GB for `.git` and the source files. How much disk space does the builder have?

kevinsawicki:
> How much disk space does the builder have?

500GB. The mac worker has two full checkouts of libcc though, mas and x64.

poiru:
@kevinsawicki Could you trigger the Linux CI for this PR?

I also rebased this on top of the `chrome56` branch just to see if it still works. If all builds are green, maybe we can just land this as part of the Chrome 56 PR?

zcbenz:
I'll test this branch locally, if we decide to take it I'll cherry pick the changes on `chrome56` branch.  And please avoid pushing any change to this branch before `chrome56` branch is merged, because it would trigger a rebuild on CI machines and development on `chrome56` branch could be interrupted for hours.

Regarding this change, one problem is that it greatly increased download size for Release builds, it was 1.3Gb, and becomes 5.0Gb after enabling `is_official_build`, this is because the compiler no longer optimizes single objects files with the whole program optimization.

The artifacts of libchromiumcontent are hosted at and downloaded directly from S3, which is very slow to access from anywhere in the world except for United States, so even with a fast home bandwidth, it can still take up to 4 hours to bootstrap Electron if the developer is not located in United States.

> Would there be any benefit from doing this since shared_library builds are supposed to only be used for local development? Improved performance at the expense of link time might not be the right tradeoff for development builds.

The `is_official_build` flag also changes behavior of certain code, so we have to be consistency for both builds, for example:
https://cs.chromium.org/chromium/src/content/app/content_main_runner.cc?q=defined%5C(official_build%5C)&sq=package:chromium&l=241&dr=C

And the whole program optimization does not affect dynamic linking as far as I know, so it shouldn't increase link time for development builds.

zcbenz:
@poiru I still get LNK1107 error on my local Windows machine:

```
codes\electron\vendor\brightray\vendor\download\libchromiumcontent\static_library\v8.lib : fatal error LNK1107: invalid or corrupt file: cannot read at 0xFF39EDB
```

Is it working fine on your machine?

poiru:
>I still get LNK1107 error on my local Windows machine:

@zcbenz Oh dear, I think I just wasted your time because I thought you got that error when *building* libcc rather than when linking Electron with libcc! I never tried the latter on Windows. Very sorry about that!

I'll give it a shot on Windows and if it does end up working for me for some odd reason. That seems unlikely based on your comment so I'll go ahead and preemptively close this.

PR #259, opened 2017-01-31
Set symbol level for shared_library and create zip with debug info.

tarruda:
Compile shared library with more debug information, and produce a .zip containing this data, which will make it simpler to debug prebuilt libchromiumcontent binaries on linux/mac. This change won't affect the libchromiumcontent.zip size, since symbols are stripped and stored separately.


tarruda:
I've changed the `create-dist` script to only extract debug symbols if `--create-debug-archive` is passed as command-line option.

zcbenz:
👍

PR #263, opened 2017-02-27
add patch to expose StreamResourceHandler

deepak1556:
This allows us to intercept the resource responses initiated with `ResourceDispatcherImpl` as streams. Pdf rendering depends on this.

https://github.com/electron/electron/pull/8435


zcbenz:
👍

PR #265, opened 2017-03-05
Add support for child and popup windows in OSR

brenca:


zcbenz:
This looks good to me, but it would take a while to merge, since there are changes to patches and we have to do a rebuild of libcc.

PR #267, opened 2017-03-09
Fix regression of drag-n-drop between different webviews

brenca:
The regression occured because of [recent changes](https://chromium.googlesource.com/chromium/src/+/7dfa247951db2ccb10b8b5276eb58a33cda4728f%5E%21/) in the chromium code (most likely related to [out-of-process iframes](http://www.chromium.org/developers/design-documents/oop-iframes)).


kevinsawicki:
@brenca I tried out this build in Electron and didn't see the issue fixed, I think the build didn't apply this patch since it was originally at the root of the repo instead of the `patches/` folder.

I've moved it into that directory and rebased, will try verifying again once the builds complete.

brenca:
@kevinsawicki Oh dang, I forgot to move that before submitting the PR, sorry for that. Another reason for me not to work at night, I guess. :D

kevinsawicki:
@brenca thanks for investigating and fixing this, verified it with the test app from https://github.com/electron/electron/issues/8667

PR #270, opened 2017-03-10
Add HeapSnapshotGenerator crash patch

kevinsawicki:
Pulls in https://codereview.chromium.org/2669393002 to prevent crashes when taking heap snapshots, https://bugs.chromium.org/p/chromium/issues/detail?id=661223

This can be dropped once Chrome 58 is upgraded to.

Refs https://github.com/electron/electron/issues/8791


paulcbetts:
:bow:, this has been trolling me

PR #272, opened 2017-03-11
Dont include widevine stub code in non component build

deepak1556:
This fixes widevine failures with release builds of Electron.

Fixes https://github.com/electron/electron/issues/8824
Ref https://github.com/electron/electron/issues/6516

Since we ask users to use the adapter and plugin from the official chrome builds, do we need to build the `widevinecdmadapter` target ?

### DIFF
--- a/chromiumcontent/build_libs.py
+++ b/chromiumcontent/build_libs.py
@@ -81,7 +81,6 @@ with open(args.out, 'w') as out:
             "third_party/sqlite",
             "third_party/sudden_motion_sensor",
             "third_party/usrsctp",
-            "third_party/widevine/cdm/widevinecdm",
             "third_party/woff2",
             "third_party/zlib",
             "tools",


deepak1556:
> Have you tested this change? Does Electron still build in both debug and release configurations on all platforms?

I have tested the release and debug builds of Electron on linux. On OSX I have just verified that the release and debug builds are created without any issues, facing some issues getting widevine plugin for OSX . So didn't test them for widevine video playback but I am sure the issue will be fixed. Would like some help building and testing on windows though, don't have access to the platform.

>  Seems like just removing widevinecdm would cause link issues.

Just curious, why do you think there will be linking issues ? Electron doesn't depend on the widevine adapter and plugin created from libcc build, we only use the generated headers for some widevine constants.

alespergl:
> Electron doesn't depend on the widevine adapter and plugin created from libcc build, we only use the generated headers for some widevine constants.

Ok then, I don't have enough info on this and was under the impression that there's code which is linked to the final product. If that is not the case, removing these object from libcc should really have no effect. Can you explain what is the effect of the change and how it fixes the mentioned problem?
Thanks.

Btw I can help with testing the Windows build.

deepak1556:
>  Can you explain what is the effect of the change and how it fixes the mentioned problem?

Pepper uses `CreateCdmInstance` as entry point for cdm creation [here](https://cs.chromium.org/chromium/src/media/cdm/ppapi/ppapi_cdm_adapter.cc?l=348), `stub_cdm.cc` contains a global `CreateCdmInstance` which overrides the value from  widevine loaded dynamically by users, hence most of the functionalities failed. This change removes it.

```
>  readelf --symbols ./electron-1.6.3/electron | c++filt | grep CreateCdmInstance
  9959: 00000000039b7e50    65 FUNC    GLOBAL DEFAULT   13 CreateCdmInstance

> readelf --symbols ./electron-1.4.14/electron | c++filt | grep CreateCdmInstance
<empty>
```

> Btw I can help with testing the Windows build.

Thanks! You can test if the shaka player widevine demo plays with the release build of Electron.

```
'use strict';                                                                   
const {app, BrowserWindow } = require('electron');                              
                                                                                
app.commandLine.appendSwitch('widevine-cdm-path',
  '/opt/google/chrome/libwidevinecdmadapter.so');

app.commandLine.appendSwitch('widevine-cdm-version',
  '1.4.8.962');

app.on('ready', () => {
  const win = new BrowserWindow({                                                                                                     
    webPreferences: {                                             
      plugins: true,
    },                                                                          
  });                                                                                                            
  const url = 'https://shaka-player-demo.appspot.com/demo/';                    
  win.loadURL(url);                                                             
});
```

deepak1556:
@alespergl the issue doesn't seem to have affected windows or mac electron release builds. Thanks to @kevinsawicki for testing it out. My conslusion seems false then, not sure why only linux builds are affected.

alespergl:
Ok, thanks for the info. Hopefully you'll find the real cause of the problem.

zcbenz:
I think it is because on Linux the symbols of widevine have been kept because of the `--whole-archive` parameter, while they have been removed on Windows and Mac. So loading a widevine plugin on Linux would fail because of duplicate symbols.

zcbenz:
Merging since it seems to work on all platforms.

PR #275, opened 2017-03-14
Streamline build output directories

alespergl:
This is to make the output directory structure more sensible in light of the new GN build. Each configuration (shared_library, etc.) has an equally named subdirectory under an architecture-based `out` directory, e.g.:
```
out-x64
├ ffmpeg
├ shared_library
└ static_library
```

### DIFF
--- a/script/build
+++ b/script/build
@@ -5,7 +5,7 @@ import os
 import subprocess
 import sys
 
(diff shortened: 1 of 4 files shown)


zcbenz:
👍

PR #278, opened 2017-03-25
Add patch to expose ScopedClipboardWriter::WriteData

poiru:
This is required to implement `clipboard.writeBuffer()` in Electron.

This patch has also been upstreamed:
https://codereview.chromium.org/2777723002

### DIFF
--- a/patches/scoped_clipboard_writer.patch
+++ b/patches/scoped_clipboard_writer.patch
@@ -1,0 +1,57 @@
+diff --git a/ui/base/clipboard/scoped_clipboard_writer.cc b/ui/base/clipboard/scoped_clipboard_writer.cc
+index 6850cd460b1d..6d652cca59b4 100644
+--- a/ui/base/clipboard/scoped_clipboard_writer.cc
++++ b/ui/base/clipboard/scoped_clipboard_writer.cc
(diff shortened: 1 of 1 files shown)


deepak1556:
Just curious, any reason why we can't use the pickle version of the function to implement the api ? As far as I understand pickle is just a container for the basic types.

poiru:
@deepak1556 If we were to store some blob using Pickle, other applications would have to deserialize the Pickle format to access the actual blob.

deepak1556:
Ah makes sense. Does enabling users to write blob data to clipboard safe ? I believe browsers avoid support for it because of security concerns. Although we don't have the same criteria as browsers but there are users embedding untrusted remote contents and they will have access to this clipboard api.

poiru:
@deepak1556 The new async clipboard API spec says:

```
The data types that untrusted scripts are allowed to write to the clipboard are limited as a security precaution. Untrusted scripts can attempt to exploit security vulnerabilities in local software by placing data known to trigger those vulnerabilities on the clipboard.
```

That is definitely a concern, but considering that Electron apps can just use a native module to write arbitrary data to the clipboard, I'm not sure if we can do anything to protect app authors from themselves in this case.

kevinsawicki:
This would unblock https://github.com/electron/electron/issues/5698 right?

poiru:
@kevinsawicki Yep!

PR #284, opened 2017-04-10
Disable external popup menus patch

gerhardberger:
This PR adds an IPC message that can be sent from the browser to the renderer to disable the use of external popup menus. This is important for offscreen rendering on macOS to be able to render popups.

Related to [#8839](https://github.com/electron/electron/pull/8839)

### DIFF
--- a/patches/disable-external-popup.patch
+++ b/patches/disable-external-popup.patch
@@ -1,0 +1,50 @@
+diff --git a/content/renderer/render_thread_impl.cc b/content/renderer/render_thread_impl.cc
+index d569350..7f5ba6e 100644
+--- a/content/renderer/render_thread_impl.cc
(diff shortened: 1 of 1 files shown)


deepak1556:
Is there a reason why the ipc cannot be implemented in Electron ? `setUseExternalPopupMenus` seems to be made available for embedders.

gerhardberger:
@deepak1556 yeah, that might be a better approach, it would be better to keep it in Electron. I will take a look at it and if I can solve it there I'll close this.

gerhardberger:
[Fixed](https://github.com/electron/electron/pull/8839/commits/2e3f85947bd12a91561d21bc3caa15a726d07c89) on Electron side.

deepak1556:
👍

PR #285, opened 2017-04-13
Backport calling EnableNonClientDpiScaling

alespergl:
This is needed to enable non-client area scaling when the screen DPI changes on Windows 10.
It can be removed after upgrade beyond Chromium 56.

### DIFF
--- a/patches/enable_non_client_dpi_scaling.patch
+++ b/patches/enable_non_client_dpi_scaling.patch
@@ -1,0 +1,110 @@
+diff --git a/ui/gfx/win/window_impl.cc b/ui/gfx/win/window_impl.cc
+index f7744687c97b..bf6994d94b1a 100644
+--- a/ui/gfx/win/window_impl.cc
++++ b/ui/gfx/win/window_impl.cc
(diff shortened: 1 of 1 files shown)


poiru:
IMO we should just add this to the manifest:
```
<dpiAwareness xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">permonitorv2, permonitor</dpiAwareness>
```

That should be simpler and will be enabled for the entire process instead of just Chromium windows (e.g. for native modules).

paulcbetts:
This is a good fix but should we just cool our heels for M58 since it's out so soon?

alespergl:
@poiru We'll need to add that to the manifest too, but it's unfortunately not enough for Windows 10 versions before the Creators Update, where calling `EnableNonClientDpiScaling` is required. I tested that scenario. Plus, this patch will come down eventually from Chromium anyway, at which point we will remove this.

@paulcbetts I'm not pushing for merging this, just making it available in case Chromium 58 gets delayed. @kevinsawicki and @zcbenz have that context, so I'll let them decide whether to merge this or not. 
:cake:

zcbenz:
It looks good to me to have this in 1.6.x. 👍

PR #280, opened 2017-04-01
Upgrade to Chrome 58

zcbenz:
- [x] Update all the patches
- [x] Compile on Linux
- [x] Compile on OS X
- [x] Compile on Windows
- [x] Generate distributions on Linux
- [x] Generate distributions on OS X
- [x] Generate distributions on Windows
- [x] Check everything works with Electron

PR #286, opened 2017-04-18
Add patch to allow notification actions

MarshallOfSound:
Working on adding inline reply support, need to unlock the `actions` array so that users can construct a new notification with

```js
new Notification(title, {
  actions: [
    type: 'text',
    title: 'Reply',
    action: 'reply'
  ]
})
```

This is just to generate the builds required to do the wiring in brightray


zcbenz:
CI machines require some changes in `chrome58` to work, let's wait until #280 gets merged.

MarshallOfSound:
@zcbenz Sorry forgot about this.  Decided that adding features through the Chromium model for Notifications was too restrictive to add all the potential options that are actually available on systems.

Stopped my work in brightray and moved it over to Electron itself in https://github.com/electron/electron/pull/9269

PR #289, opened 2017-05-09
Stop using experimental allocator shim on win32

zcbenz:
It is causing electron.exe and node.dll to use different allocators, this is fine in Chromium becuase everything has been compiled into the same exe, while in Electron lots of components have been put into node.dll and we have to make sure electron.exe and node.dll use the same allocator.

This is to fix the crash in https://github.com/electron/electron/pull/9116.


kevinsawicki:
👍 Retriggered the linux builds and they are passing

PR #292, opened 2017-05-13
arm64 build

kolach:
option to build arm64 version

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,3 +43,13 @@ $ ./chrome/installer/linux/sysroot_scripts/install-debian.wheezy.sysroot.py --ar
 $ cd -
 $ ./script/build -t arm
 ```
+
+
+### Building for ARM64 target
+
+```bash
+$ ./script/bootstrap
+$ ./script/update -t arm64
+$ ./script/build -t arm64
+$ ./script/create-dist -t arm64
+```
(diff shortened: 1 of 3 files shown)


kolach:
My aim is to contribute into arm64 Electron version. And libcromiumcontent component is the first step to make it possible.

I can already tell that I successfully used the compiled libcromiumcontent arm64 version to build Electron using params of the bootstrap.py script.

```bash
$ ./script/bootstrap.py -v --target_arch=arm64 --
libcc_source_path=../libchromiumcontent/src
libcc_static_library_path=../libchromiumcontent/dist/main/static_library --libcc_shared_library_path=../libchromiumcontent/dist/main/shared_library --
```

Of cause I also made several modification in several Electron scripts. But that will go as a PR to electron repository.

kolach:
@zcbenz Can you explain me what project should I use to make downloadable sysroot for Electron. For my arm64 experiment I used the sysroot scripts inside libcromiumcontent. I only had to add libnotify library to it. But I do not think it's a right way. Do you use some dedicated project for that purpose?

zcbenz:
@kolach The sysroot images were generated in https://github.com/electron/debian-sysroot-image-creator, we need to add arm64 images.

PR #293, opened 2017-05-14
backport fix for accented letter input resulting double letters in webview

deepak1556:
Backports https://codereview.chromium.org/2689153002/ which fixes a long standing IME bug, can be removed with chromium 59 upgrade.

Fixes https://github.com/electron/electron/issues/3379


kevinsawicki:
Built and tested this locally and the IME not works as expected, thanks for this @deepak1556 👍 ⌨️ 

Will merge this and then wait for all the libcc builds to finish before upgrading Electron.

PR #296, opened 2017-05-19
Add webgl_context_attributes patch

miniak:
This patch adds `powerPreference` to `WebGLContextAttributes`, which allows to optionally prefer using the integrated GPU for WebGL contexts:
https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.2.1

Closes https://github.com/electron/electron/issues/9528

### DIFF
--- a/patches/webgl_context_attributes.patch
+++ b/patches/webgl_context_attributes.patch
@@ -84,7 +84,7 @@ index 9a48073..5196a53 100644
      boolean failIfMajorPerformanceCaveat = false;
  };
 diff --git a/third_party/WebKit/public/platform/Platform.h b/third_party/WebKit/public/platform/Platform.h
-index 8c078e7..0779d4a 100644
+index 2cd7334..98ecb84 100644
 --- a/third_party/WebKit/public/platform/Platform.h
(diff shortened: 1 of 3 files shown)


miniak:
I am having an issue with `worker_context_will_destroy.patch` modifying `third_party/WebKit/public/platform/Platform.h`, which my patch is also changing. If my patch is named `webgl_prefer_integrated_gpu.patch`, it's applied before the other patch, which then fails complaining that `Platform.h` has been modified.

miniak:
@kevinsawicki, can you please have a look? Thanks

kevinsawicki:
> I am having an issue with worker_context_will_destroy.patch modifying third_party/WebKit/public/platform/Platform.h, which my patch is also changing

I updated the worker patch to be applicable after this one and dropped the `zzz_` prefix.

Tested this out on macOS and verified the activity monitor showed the right `Requires High Perf GPU` call when doing: 

```js
document.createElement('canvas').getContext('webgl', {powerPreference: "low-power"});
```

Thanks for this @miniak 👍 :shipit:

PR #297, opened 2017-05-19
Add inspector number parsing patch

kevinsawicki:
Backports https://codereview.chromium.org/2866213002 and  https://codereview.chromium.org/2846673005 which have been fixed in Chrome 59.

Refs https://bugs.chromium.org/p/chromium/issues/detail?id=712610
Closes https://github.com/electron/electron/issues/8876


kevinsawicki:
Verified this on the macOS build that passed 🚢

PR #300, opened 2017-05-22
Backport macOS IME backspace fix

kevinsawicki:
Backports https://codereview.chromium.org/2901443002/

Refs https://bugs.chromium.org/p/chromium/issues/detail?id=714771
Fixes https://github.com/electron/electron/issues/9173


kevinsawicki:
Verified this fixed the issue with the example app from https://github.com/electron/electron/issues/9173 on macOS using Katakana layout.

PR #304, opened 2017-05-23
Add patch that removes setting If-Match headers

kevinsawicki:
Backports https://codereview.chromium.org/2714583002 to Chrome 56.

Refs https://bugs.chromium.org/p/chromium/issues/detail?id=689989

This does not affect Chrome 58 (Electron 1.7.x) so it is only being backported to Chrome 56 (Electron 1.6.x)

/cc @deepak1556 @CharlieHess


kevinsawicki:
Tested this out on macOS using http://clips.vorwaerts-gmbh.de/big_buck_bunny.webm and scrubbing quickly before the video fully loads, appears to be working now as expected, below is the network timeline for 1.6.10 vs. this branch

### 1.6.10

screen shot 2017-05-24 at 10 27 21 am


### This branch

screen shot 2017-05-24 at 10 27 56 am

PR #303, opened 2017-05-23
Add RenderWidget patch for macOS IME backspace fix

kevinsawicki:
Backports #300 to Chrome 56 (Electron 1.6.x).

/cc @CharlieHess @deepak1556

PR #305, opened 2017-05-24
Use vendored boto

kevinsawicki:
This was done in https://github.com/electron/libchromiumcontent/pull/280 but needs to be done on the `electron-1-6-x` branch as well since the new CI machines don't have boto installed globally.

This pull request just cherry-picks the commits from that pull request.


kevinsawicki:
The linux arm build succeeded and uploaded so I'm going to merge this before the other builds complete to get #304 and #303 going.

PR #306, opened 2017-05-25
Add patch to revert https://codereview.chromium.org/2745153002/

poiru:
This was suggested in
https://bugs.chromium.org/p/chromium/issues/detail?id=725970#c5 to fix
electron/electron#9578.

Reverts https://codereview.chromium.org/2745153002


kevinsawicki:
Verified it on macOS 👍

PR #307, opened 2017-05-26
Set pipefail to catch cibuild errors

kevinsawicki:
Linux builds are still green when failing without since `grep` is used to filter out thousands of ignored warnings.


kevinsawicki:
Got a proper failure now on the arm build 🚢

PR #309, opened 2017-05-30
Apply Chromium patches using git and sync via gclient

### LINKED ISSUE #282: CI builds may not be idempotent
@kevinsawicki CI builds currently reuse `libchromiumcontent/src` if the Chromium version was the same as the last build. This seems super dangerous to me because patches applied by one build will stay applied as long as the Chromium version doesn't change. Should we [refetch the tarball](https://github.com/electron/libchromiumcontent/pull/281/commits/429619a5c94a1dce1c472a9315e1a1fc51e9914a) each time?

kevinsawicki:
This pull request switches the build from downloading a chromium src tarball to using `gclient` to sync chromium.

* Makes it easier to apply patches via `git apply`
* Patches can be reverted using `git reset` or `gclient sync --reset`.
* Patches can be created without having a separate chromium src repo cloned locally

Fixes #282


kevinsawicki:
@zcbenz are you okay with this approach?

It does use more disk space but would make switching between Chrome versions while offline possible and quicker since it would only need to checkout a different revision instead of downloading a new tarball each time.

kevinsawicki:
Although this is working currently, I'm going to take a look at using `gclient` for this instead since the current approach in this PR would make a clone of electron/electron significantly larger since it would clone the chromium submodule even though most people just use the pre-built libcc assets for electron builds. Using `gclient` would only cause the chrome code to be synced when building this repository.

tonyganch:
Have you considered replacing patches with actual commits in chromium mirror?
We've spent a day updating patches for Ch59, and most of the changes were in line numbers and surrounding lines (3 lines before/after the actual diff).
I believe having commits instead of patch files will remove that part of work and leave only changes in diffs themselves.
What do you think of that?

kevinsawicki:
> I believe having commits instead of patch files will remove that part of work and leave only changes in diffs themselves.

The patches should apply easier using `git apply` than the `python-patch` module so they shouldn't need to be updated on each chrome release unless true conflicts exist. I'll verify this on 59 today to double check.

> Have you considered replacing patches with actual commits in chromium mirror?

Yeah, I've thought about this a bit, it would require forking several repos since we patch chrome, v8, ffmpeg, icu, and skia. It would also require maintaining commits that alter the root `DEPS` file to pull from the forks instead of from the main chrome repos.

kevinsawicki:
The builds are now passing, I'm going to merge this as-is since it enables predictable builds and prevents the issues mentioned in #282 regarding patches across different branches.

If we decide to fork chrome instead of patching it that work can build on top of this pull request since we can re-use the `gclient` workflow and just point it to the fork instead of the main chromium repo.

@tonyganch @cifratila perhaps we can explore the fork approach as part of the chrome 59 upgrade to see how much easier it might be compared to the patches.

PR #310, opened 2017-06-11
linux appIndicator submenu fix

dregenor:
this fix for issue with disabled submenu if in same position in main menu item disabled
https://github.com/electron/electron/issues/9495


kevinsawicki:
@dregenor looks like this is targeting the 1-6-x branch, does it happen in 1.7.x as well? Should it be fixed in both? Thans.

dregenor:
@kevinsawicki  yes this issue reproduced also for 1-7-x
https://chromium.googlesource.com/chromium/src.git/+log/58.0.3029.110/chrome/browser/ui/libgtkui/app_indicator_icon_menu.cc - last changes 4 months ago

kevinsawicki:
Thanks for this @dregenor

PR #312, opened 2017-06-15
Use dynamic CRT in the Windows static library build

alespergl:
This addresses [electron#9443](https://github.com/electron/electron/issues/9443)


alespergl:
Since I got a 👍 from @zcbenz and no complaints, I will go ahead and merge this. I will then follow up by merging https://github.com/electron/electron/pull/9764

PR #314, opened 2017-06-23
linux appIndicator submenu fix

dregenor:
this fix for issue with disabled submenu if in same position in main menu item disabled
electron/electron#9495
port from 1-6-0


kevinsawicki:
Thanks for this @dregenor 👍

PR #318, opened 2017-07-05
Revert "Backport macOS IME backspace fix"

kevinsawicki:
Looks like this introduced electron/electron#9709 and so reverting until Chrome 59 can be tested/verified.

Reverts electron/libchromiumcontent#300


kevinsawicki:
Can re-apply once Chrome 59 lands and include https://codereview.chromium.org/2929543002/ as well.

PR #319, opened 2017-07-11
Apply patches fix for cases when src is not a git repo

cifratila:
This fix is compatible to any setup you choose for your LIBCC build. 

If you don't want to use `gclient`, then the `src` folder will not be a git repo and the current approach will fail applying patches (actually skipping them). The reason for this might be that you are not allowed to fetch dependencies from outside of your network and everything needs to be "cooked" in house.


kevinsawicki:
Thanks @cifratila 👍

PR #322, opened 2017-07-17
Disabled ASSERT which causes some Electron tests to fail

alespergl:
Follow-up to enabling debug mode.


alexeykuzmin:
@kevinsawicki Can you please run Linux builds for this PR?

alexeykuzmin:
We should probably create a libcc branch with all the ASSERTS/DHECKS enabled
to be able to fix Electron tests that fail on them.
@alespergl Can you please do that after this PR is merged?

kevinsawicki:
> Can you please run Linux builds for this PR?

Just triggered them 👍

PR #313, opened 2017-06-19
Upgrade to Chromium 59

alexeykuzmin:
This pr is created as a replacement for the #311.

/cc @tonyganch @kevinsawicki @zcbenz

PR #324, opened 2017-07-18
Use Docker for Linux builds

kevinsawicki:
Allows us to easily update the distro and config for the prebuilt Linux libraries.

Will merge into #313 once the build goes green here so those builds can successfully complete.

Copies in https://chromium.googlesource.com/chromium/src/+/lkcr/build/install-build-deps.sh and tweaks it so it can be run in Docker.

Refs https://github.com/electron/libchromiumcontent/pull/313#issuecomment-311901269

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+src
+dist
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,3 @@
 src
 dist
+.git
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 /vendor/binutils-aarch64
 .gclient_entries
 _gclient_src_*
+s3credentials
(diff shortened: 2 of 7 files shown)


kevinsawicki:
The 64 bit succeeded but 32 failed due to https://bugs.chromium.org/p/chromium/issues/detail?id=729532 and arm failed with these errors:

```
../../build/linux/debian_jessie_arm-sysroot/usr/include/freetype2/config/ftconfig.h:417:5: error: 'register' storage class specifier is deprecated and incompatible with C++1z [-Werror,-Wdeprecated-register]
    register FT_Int32  t, t2;
    ^~~~~~~~~
../../build/linux/debian_jessie_arm-sysroot/usr/include/freetype2/config/ftconfig.h:417:5: error: 'register' storage class specifier is deprecated and incompatible with C++1z [-Werror,-Wdeprecated-register]
    register FT_Int32  t, t2;
    ^~~~~~~~~
2 errors generated.
``

PR #326, opened 2017-07-23
Enabled ANGLE and Skia debug checks

alespergl:
Please use builds from this branch to fix debug check violations in Electron.

This particular branch only enables debug checks in ANGLE and Skia libraries, so that it's possible to focus on violations specifically in those libraries.

There's another branch which enables *all* debug checks, and presents a greater challenge for fixing bugs:
https://github.com/electron/libchromiumcontent/pull/327

**Tip:** Symbol files for libcc are stored at the same location as the published `libchromiumcontent.zip` file, only the file is named `libchromiumcontent-dbg.zip`

PR #327, opened 2017-07-23
Enabled all debug checks

alespergl:
Please use builds from this branch to fix debug check violations in Electron.

This branch enables *all* debug checks including V8 and general Chromium ones, and as such presents the greatest challenge because Electron triggers those checks often.

There's another branch which enables only some debug checks whose violations should be easier to fix:
https://github.com/electron/libchromiumcontent/pull/326

**Tip:** Symbol files for libcc are stored at the same location as the published `libchromiumcontent.zip` file, only the file is named `libchromiumcontent-dbg.zip`


alespergl:
Rebased on current master.

alexeykuzmin:
@alespergl Do you still need this PR?

deepak1556:
`Angle`, `Skia` and `v8` are the only components that have DCHECK disabled apart from the existing `002-dcheck.patch`. I think we can use https://github.com/electron/libchromiumcontent/pull/326 and close this one. Trying to fix with all of them enabled at once will be difficult.

alespergl:
Let's leave them both open for now as a reminder that there's still work to be done. I'll rebase them on master as soon as I have a moment.

ckerr:
@alespergl Do you still need this PR?

alexeykuzmin:
Since there's no response from @alespergl and the libcc repo is going to be archived soon, I'm going to close this PR.

PR #328, opened 2017-07-25
Add `script/patch.py` to use during Chromium upgrades

alexeykuzmin:
**Motivation**:
`apply-patches` script is perfectly fine when patches could be applied without any errors,
but during upgrades to a new Chromium version it's almost useless.
`patch.py` should make work with failing patches more versatile and easier.

### DIFF
--- a/script/apply-patches
+++ b/script/apply-patches
@@ -4,6 +4,8 @@ import os
 import subprocess
 import sys
 
+from patch import apply_patches_from_directory
+
 
 SOURCE_ROOT = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
 PATCHES_DIR = os.path.join(SOURCE_ROOT, 'patches')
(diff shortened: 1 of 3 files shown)


alexeykuzmin:
Are the docstring and arguments description not enough?

```
"""
Usage: patch -h

Use this script to selectively apply and revert patches.
It is most useful during upgrades to a new Chromium version.
"""
```

$ ./patch -h
```
usage: patch [-h] [--directory DIRECTORY] [--file_path FILE_PATH] [--force]
             --repo_root REPO_ROOT [--reverse]

Apply patches to a git repo

optional arguments:
  -h, --help            show this help message and exit
  --directory DIRECTORY
                        Path to a directory with patches. If present,
                        --file_path is ignored.
  --file_path FILE_PATH
                        Path to a patch file.
  --force               Try apply all the patches even if some of them fail.
  --repo_root REPO_ROOT
                        Path to a repository root folder.
  --reverse             Reverse patches.
```

cifratila:
@alexeykuzmin I was thinking more of adding it maybe for others here: https://github.com/electron/electron/blob/master/docs/development/upgrading-chrome.md

alexeykuzmin:
> @alexeykuzmin I was thinking more of adding it maybe for others here: https://github.com/electron/electron/blob/master/docs/development/upgrading-chrome.md

The whole manual should be updated =/ I'm going to do that someday.
Anyway, it's in a different repo, so definitely not in this PR.

alexeykuzmin:
Any other comments or suggestions?

zcbenz:
👍

PR #330, opened 2017-08-01
Improve 1 6 build output

zeke:
This PR cherry-picks two commits to make debugging 1.6 builds easier.

- https://github.com/electron/libchromiumcontent/pull/316 Surface boto process errors 
- https://github.com/electron/libchromiumcontent/pull/307 Set pipefail to catch cibuild errors


script/cibuild-libchromiumcontent-linux
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 set -e
+set -o pipefail
```
poiru: You can replace these two `set` lines with `set -o errexit -o pipefail` (arguably all of our shell scripts should start with that).

poiru: Derp, I missed the part about cherry-picking. Nevermind!

PR #331, opened 2017-08-01
Backport fix for flickering web contents on large monitors

deepak1556:
Backports https://chromium-review.googlesource.com/c/575445/

Refs https://bugs.chromium.org/p/chromium/issues/detail?id=695427

Fixes https://github.com/electron/electron/issues/9413

I wasn't able to reproduce the issue, so can't verify if this actually fixes the issue. Backporting it so that anyone interested can verify with the libcc build.

/cc @bpasero

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,239 @@
+version: 2.0
+jobs:
+  libchromiumcontent-linux-x64:
+    docker:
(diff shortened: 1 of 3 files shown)


bpasero:
Thanks, very interested in seeing this land. Can also verify easily.

PR #333, opened 2017-08-04
Disabled debug allocation as it causes some Electron tests to fail

alespergl:
Further CI tests fail (on Linux) due to a memory corruption flagged in the debug mode. This disables the debug allocation framework to prevent those failures.

### DIFF
--- a/patches/build_gn.patch
+++ b/patches/build_gn.patch
@@ -1,8 +1,15 @@
 diff --git a/base/allocator/BUILD.gn b/base/allocator/BUILD.gn
-index 8cdb06161f5c..285461d3381d 100644
+index 8cdb06161f5c..6f938d6be67b 100644
 --- a/base/allocator/BUILD.gn
 +++ b/base/allocator/BUILD.gn
-@@ -16,6 +16,7 @@ declare_args() {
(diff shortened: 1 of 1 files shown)


zcbenz:
I hope it could fix the problem in https://github.com/electron/electron/pull/9946#issuecomment-320216716.

alexeykuzmin:
Could we just set `enable_debugallocation=false` in build config (or `script/update`) instead of changing the code?

alespergl:
Oh, right, we probably could do that. I didn't realize it was in the `declare_args` section.

alexeykuzmin:
> Oh, right, we probably could do that.

I think we can change it later ) It would take too much time to wait for all the builds again.

alespergl:
@alexeykuzmin Why the rush? I would have preferred if you gave me a chance to incorporate your feedback before merging. We still could have used the existing version for validating that it helps with the tests.

alexeykuzmin:
> Oh, right, we probably could do that. I didn't realize it was in the declare_args section.

@alespergl Sorry, I didn't understand from your reply that you were going to make changes in this PR.

PR #335, opened 2017-08-07
Upgrade to Chromium 61

alexeykuzmin:
### Curent state of the branch:
Chromium 61.0.3163.100.

### TODO:
- [x] (Node 8.7.0) Fix compilation error on Linux https://circleci.com/api/v1.1/project/github/electron/libchromiumcontent/788/output/104/0?file=true
- [x] [_Weird compilation errors, skip for now_] Enable Jumbo build?
- [x] Use 61.0.3163.100
- [x] Rebase onto the latest master (after 02/10/2017)
- [x] Remove RCE fix patch after the rebase
- [x] Fix ARM 64
- [x] Remove `download_gold` from `script/update` after the rebase
- [x] Fix Win compilation.
- [x] Change target branch to *master* once Chromium 59 is merged.
- [x] [_Yes, it is._] Check if `patches/unused_local_var.patch` is still needed after rebase on the latest master.
- [x] Win32 `IOError: [Errno 2] No such file or directory: 'C:\\e\\workspace\\libcc-win-ia32\\src\\out-ia32\\shared_library\\obj\\third_party\\webrtc\\base\\rtc_base_cc.pdb'`
- [x] Use Chromium 61.0.3163.59
- [x] Rebase onto the latest master (after 23/08/2017)


script/update
```diff
@@ -234,7 +234,7 @@ def run_gn(target_arch, defines):
   for component in COMPONENTS:
     args = 'import("//chromiumcontent/args/{0}.gn") target_cpu="{1}"'.format(component, target_cpu)
     if sys.platform in ['win32', 'cygwin']:
-      args += ' use_experimental_allocator_shim=false'
+      args += ' use_allocator_shim=false'
```
alexeykuzmin: just a reference
https://codereview.chromium.org/2902043007

alexeykuzmin:
```
if target_arch =='arm64':
  args += ' use_system_freetype=true'
```

@zcbenz that [Chromium issue](https://bugs.chromium.org/p/chromium/issues/detail?id=274030) says

> this CL introduces a
switch use_system_freetype which can be used to keep the old linking
configuration at the expense of risking introduction of text rendering
and security regressions if the system FreeType is too old.

Are we sure that system freetype we use isn't "too old"? )

alexeykuzmin:
> Zipping shared_library builds...
Zipping static_library builds...
+ script/upload
multipart upload requested but not capable

@jkleinsc, it seems like we have some trouble at least on Mac on Jenkins.
http://208.52.191.140:8080/job/libchromiumcontent-mas/669/console

zcbenz:
> Are we sure that system freetype we use isn't "too old"? )

Linux distributions only started to have arm64 ports in recent years, so I think the system freetype is usually new on them.

Having said that, I believe root problem is they did not update arm64 sysroot image for the freetype build changes, and we can revert this once they fixed it.

zcbenz:
Regarding the multipart problem, I took a different approach in https://github.com/electron/libchromiumcontent/pull/361.

alexeykuzmin:
> > Are we sure that system freetype we use isn't "too old"? )

> Linux distributions only started to have arm64 ports in recent years, so I think the system freetype is usually new on them.

If it works for GitHub then I'm ok wit this ) @zcbenz, thanks for the fix!

alexeykuzmin:
> Regarding the multipart problem, I took a different approach in #361.

@zcbenz, @jkleinsc 
If you guys decide that approach in #361 works better then I'll remove all upload related commits from this branch and make a rebase, don't bother yourself doing that )

jkleinsc:
@alexeykuzmin I merged in #361 and I'm happy with the approach @zcbenz took, so let's go that route.

alexeykuzmin:
@alespergl 
It seems like we need `use_system_freetype = true` even for "ffmpeg" build.

alexeykuzmin:
bb33dd686a8d659fc0547098ff8b6986f12e944c has all green builds except **libchromiumcontent-osx** which wasn't run at all

zcbenz:
Close and reopen to restart the CI.

zcbenz:
I'm rebasing this branch to fix the conflict introduced by the mips64el PR. The build state is not supposed to be affected by the mips64el changes.

PR #336, opened 2017-08-08
Add OnCreateSiteInstanceForNavigation

MarshallOfSound:
This will give us access to the `SiteInstance` we need to fix https://github.com/electron/electron/pull/10204

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,13 +70,13 @@ Asset URLs have the following format:
 
 Builds exist for the following platform/arch pairs:
 
-- `osx/x64`
-- `mas/x64`
-- `win/ia32`
-- `win/x64`
+- `linux/arm`
 - `linux/ia32`
 - `linux/x64`
-- `linux/ar`
+- `mas/x64` (though `osx/mas` would be a more accurate name)
+- `osx/x64`
+- `win/ia32`
+- `win/x64`
 
 The Linux machines only build for pushes by GitHub employees, so PRs 
 from third parties have to have their linux builds triggered manually.
--- a/README.md
+++ b/README.md
@@ -53,3 +53,54 @@ $ ./script/update -t arm64
 $ ./script/build -t arm64
 $ ./script/create-dist -t arm64
 ```
+
+## Releases
+
+There is no formal release process for libchromiumcontent, as release artifacts
+are created as a byproduct of CI. When a build is successful, its compiled
+asset is automatically uploaded to S3. These assets are later downloaded as 
+part of Electron's bootstrap script.  These files are about 4GB, so the 
+bootstrap task takes a while to run.
+
+Asset URLs have the following format:
+
+```js
+`https://s3.amazonaws.com/github-janky-artifacts/libchromiumcontent/${platform}/${commit}/libchromiumcontent.zip`
+```
+
+Builds exist for the following platform/arch pairs:
+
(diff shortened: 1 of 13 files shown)


zeke:
Is there a document somewhere that describes how to bring in these patches?

MarshallOfSound:
@zeke Don't think so

zeke:
I wouldn't even know where to start. Can you share some pointers and I can work towards getting a basic doc together?

MarshallOfSound:
@zeke The way I normally generate *new* patch files is to modify the file in `src` then simply pipe the output of `git diff path/to/file` to `patch_name.patch`.  Then just put that patch file in the `patches` folder.

If a file already has a patch it gets trickier, I normally just edit the existing patch file but wouldn't recommend doing that unless you know what's going on as you'll probably just break the patch file 😄

alexeykuzmin:
@zeke Most of the patches represent a certain "feature", set of changes that add a new or change existing Chromium behaviour in some way.

To add a new patch I would apply all existing patches via `/script/apply-patches` and stage all the changes via `git add .`, then make changes for a new "feature". After that it's easy to collect a diff across the whole repository via simple `git diff` and put it into a `new.patch` file.

But there is a catch. Some patches contain changes to the same files, it means that the order they are applied matters. And if a new patch contains changes for the same files as the others it might break patches that would be applied after the new patch.
We could avoid these "conflicts" if we were able to always add patches to the end of the list.
I have created an issue to do this – #339.

zeke:
Awesome. Can you distill that into a happy-path list of commands that assumes #339 has already happened?

Here's my bad attempt:

```sh
# edit files in src, then...
./script/apply-patches
git diff path/to/file > patches/patch_name.patch
```

alexeykuzmin:
@zeke something like this:
```bash
$ ./script/update  # to get Chromium in `src/` and apply existing patches to it,
    # no need to do it again if it has already been done
$ cd src  # or to the root of a repo you are going to change code in
$ git add .  # stage existing changes to make `git diff` useful
$ # make any code changes you like
$ git diff > path/to/the/patches/xxxx-meaningful-name.patch  # where `xxxx` is greater
    # than any existing patch index, e.g. '0052' if the last patch in the folder
    # is named '0051-some-other.patch'.
```

alexeykuzmin:
And btw, `patches/` folder itself contains patches only for the files in the `src/` that belong to the Chromium repo, its subfolders (e.g. `patches/v8/` or `patches/third_party/skia/`) contain patches for the subfolders of the `src/` (`src/v8/` and `src/third_party/skia/` respectively) that are **independent** repos.
If one need to make changes in two different repos they **must** create two patch files.

alexeykuzmin:
@MarshallOfSound list of the commits looks weird, there is a commit that obviously was in the master once. It looks like someone has made a force push into the master.

/cc @jkleinsc

jkleinsc:
@alexeykuzmin that was my bad.  I thought I was on a different branch than master.

zeke:
After force-pushing to electron/electron master a few times myself, I enabled branch protection on it a while back.

I have now enabled it for this repo too. 🔒 

screen shot 2017-08-09 at 9 38 05 pm

jkleinsc:
Closing in favor of #340

PR #338, opened 2017-08-08
Patch the Windows build toolchain to generate unique PDB names

enlight:
When the PDB files generated by the `static_library` build are packaged for distribution they are all copied to a single folder, some of the PDB files have identical names so they end up overwriting each other. The missing PDB files cause linker warnings when building Electron in Release mode, and make it more difficult
to debug release builds.

This patch modifies the PDB naming convention for the `static_library` build configuration to ensure PDB names are unique. For example, instead of generating `obj/ui/base/base_cc.pdb` the build will now generate `obj/ui/base/obj_ui_base_base_cc.pdb`.

@alespergl suggested an alternative approach (using the `/Z7` compiler flag) that might be worth experimenting with, it may end up being less hacky and more performant than this patch.

### DIFF
--- a/patches/build_toolchain_win_patch.patch
+++ b/patches/build_toolchain_win_patch.patch
@@ -1,0 +1,82 @@
+diff --git a/build/toolchain/win/BUILD.gn b/build/toolchain/win/BUILD.gn
+index dd1017f..d15706b 100644
+--- a/build/toolchain/win/BUILD.gn
++++ b/build/toolchain/win/BUILD.gn
+@@ -129,12 +129,19 @@ template("msvc_toolchain") {
+     tool("cc") {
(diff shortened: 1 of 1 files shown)


zcbenz:
@enlight Can you rebase this on master and then push it to a branch of this repo? So I can run CI jobs for it.

enlight:
@zcbenz OK, rebased (after some effort since someone force pushed master), and created a branch in this repo.

enlight:
Let me know if you want me to create a new PR for https://github.com/electron/libchromiumcontent/tree/missing-pdb-fix since this PR is stuck on my fork.

enlight:
Closed this one in favor of #341.

PR #340, opened 2017-08-10
Add OnCreateSiteInstanceForNavigation

MarshallOfSound:
This is attempt 2 because commit's went weird


MarshallOfSound:
@zcbenz These builds are all green and with my changes in https://github.com/electron/electron/pull/10221 the tests that were failing are green as well.  Can we merge this to get the ball rolling on getting this into Electron 😄

zcbenz:
This looks good to me. /cc @alexeykuzmin This PR is going to affect chrome 59 and 61 updates.

PR #342, opened 2017-08-10
Move Linux CI to CircleCI

jkleinsc:
Moving Linux CI to CircleCI so that we can run parallel builds and also have community visibility into Linux builds/failures as well as give maintainers the ability to restart builds themselves.


zeke:
Look at all those green CircleCI checks!

zeke:
💚

Issue #339, opened 2017-08-09
Add indexes to patch files


alexeykuzmin:
```
accelerator.patch       →    0001-accelerator.patch
allow_new_privs.patch   →    0002-allow_new_privs.patch
...
```

I will allow us to always add new patches to the end of the list.

I guess we should do it only after the Ch59 branch is merged to avoid delays.
Renames will cause conflicts in the Ch61 branch, but I don't think it's a big deal.

/cc @tonyganch

zeke:
Do these indexes need to be added by file date or name or anything?

MarshallOfSound:
@zeke Just in order the patches were created, patches are applied in alphabetical order so doing this enforces an order while allowing us to name them correctly (instead of coming up with a patch name that just happens fits in the right slot alphabetically) 😆

alexeykuzmin:
@zcbenz can it lead any undesirable side effects?

zcbenz:
> @zcbenz can it lead to any undesirable side effects?

I don't think it would have undesirable side effects.

Having multiple patches modifying one file makes it very hard to update the patches for Chrome updates, you have to generate diffs carefully in right sequence and it is definitely not an easy task.

In the past I just avoided it at all, when a new patch was modifying a file already patched, I would merge the patch into other patches, or create a new patch that have all the changes to one file.

alexeykuzmin:
> Having multiple patches modifying one file makes it very hard to update the patches for Chrome updates, you have to generate diffs carefully in right sequence and it is definitely not an easy task.

It should be much easier now since we use actual Chromium repo and not a snapshot.

zcbenz:
> It should be much easier now since we use actual Chromium repo and not a snapshot.

It is still very frustrating, however if you are fine with it then I'm totally OK with this approach, since it is you doing the heavy update now 😄 .

PR #343, opened 2017-08-11
document how to add a patch

zeke:
A followup to https://github.com/electron/libchromiumcontent/pull/336

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,37 @@ $ ./script/build -t arm64
 $ ./script/create-dist -t arm64
 ```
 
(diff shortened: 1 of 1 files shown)


zeke:
PS this shouldn't be merged until https://github.com/electron/libchromiumcontent/issues/339 is fixed.

zeke:
> Can you also add a note about patches for the repos other than Chromium

Done.

zeke:
@zcbenz this is good to go now that #339 has landed.

PR #345, opened 2017-08-15
Re-enabled debug mode, and disabled tcmalloc for shared_library build

alespergl:
This PR re-introduces debug mode for the shared_library configuration.

It was disabled because there were issues with it when Electron tests were executed as part of the Linux CI build - the tests hung inexplicably. It turns out that the thing that causes the hangs is the tcmalloc allocator library, which (when run inside Docker) deadlocks while trying to allocate memory.

The way tcmalloc works is by replacing the `malloc` family of APIs normally provided by the C runtime library. Supposedly on Linux, tcmalloc has better performance compared to the default, which is why Chromium uses it. It's not used on other platforms.

I was unable to explain why the deadlock condition occurs in Docker and not elsewhere. Nevertheless, I decided to just disable tcmalloc for the shared_library build, since it is only used on Linux, and it shouldn't be a problem to use the built in glibc `malloc` for shared_library, which is Electron's debug configuration.

### DIFF
--- a/chromiumcontent/args/shared_library.gn
+++ b/chromiumcontent/args/shared_library.gn
@@ -1,8 +1,9 @@
 root_extra_deps = [ "//chromiumcontent:chromiumcontent" ]
 is_electron_build = true
 is_component_build = true
-is_debug = false
+is_debug = true
 symbol_level = 2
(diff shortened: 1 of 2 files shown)


alespergl:
@jkleinsc could we produce builds for this PR in order to validate the changes on CI?

Thanks 🙇

jkleinsc:
@alespergl builds are on their way.

PR #348, opened 2017-08-22
Add a prefix for the patch added in PR #347

enlight:
All patches should be prefixed now.


alexeykuzmin:
@enlight Thank you!

poiru:
Thanks!

PR #350, opened 2017-08-23
Break builds into static and shared builds

jkleinsc:
Breaking builds into separate static and shared builds should decrease build time

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -415,30 +415,30 @@ workflows:
       - libchromiumcontent-linux-x64-static-build
       - libchromiumcontent-linux-x64-shared-build
       - libchromiumcontent-linux-x64-upload:
-        - requires:
-              - libchromiumcontent-linux-x64-static-build
-              - libchromiumcontent-linux-x64-shared-build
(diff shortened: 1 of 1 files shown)


zcbenz:
Our internal Linux CI failed to build Electron after switching to circle CI:

```
+ apt-get update
Reading package lists...
[91mW: chmod 0700 of directory /var/lib/apt/lists/partial failed - SetupAPTPartialDirectory (1: Operation not permitted)
E: Could not open lock file /var/lib/apt/lists/lock - open (13: Permission denied)
E: Unable to lock directory /var/lib/apt/lists/
[0mError executing command, exiting
Removing intermediate container 21da1fc33727
The command '/bin/sh -c curl -sL https://deb.nodesource.com/setup_6.x | bash -' returned a non-zero code: 1
```

Do you have an idea of it?

jkleinsc:
@zcbenz it has to do with the fact that I changed the Dockerfile here on libchromiumcontent.  I have a fix in the https://github.com/electron/electron/tree/re-enable_debug_mode branch.

PR #361, opened 2017-09-24
Build static build as an official build

### LINKED ISSUE #358: Set `is_official_build=true` for static_library builds
@zcbenz, do we need it for "ffmpeg" config?

- [ ] Check that Electron can be built and works well with the new libcc.

alexeykuzmin:
Fixes #358.


zcbenz:
The `025-no_stack_dumping.patch` can be removed now.

alexeykuzmin:
@zcbenz, do we need it for "ffmpeg" config?

zcbenz:
I think we should, so we can minimize differences of compiler settings.

alexeykuzmin:
@zcbenz 

> The `025-no_stack_dumping.patch` can be removed now.

Will it be ok if we remove it and keep build shared_library without `is_official_build=true`?

zcbenz:
> Will it be ok if we remove it and keep building shared_library without is_official_build=true?

Oh it would not be ok, maybe we should also do is_official_build=true for shared library build, so we have minimal behavior differences?

alexeykuzmin:
As far as I understand all developer builds of Chromium are built without `is_official_build=true`,
and in Yandex we used to build locally without it. And I have never seen any problems because of that.
Are you sure we can't have the same for libcc?

cifratila:
@alexeykuzmin @zcbenz 

I was trying to enable the flag also in our internal builds and faced errors with the llvm gold library on Linux x64, seems also here we face the same issues.

Any ideas on how to fix it?

alexeykuzmin:
@cifratila 
Yeah, I've seen it. Have no idea how to fix it though.

zcbenz:
> As far as I understand all developer builds of Chromium are built without is_official_build=true,
> and in Yandex we used to build locally without it. And I have never seen any problems because of that.
> Are you sure we can't have the same for libcc?

I did not know that, lets keep the same with Chromium then.

zcbenz:
I'm rebasing this branch to fix the conflict file.

zcbenz:
I added a step to download the gold plugin, but note that Chromium has removed the gold plugin in later versions, so we need to remove this step in Chrome 61 branch.

alexeykuzmin:
> but note that Chromium has removed the gold plugin in later versions, so we need to remove this step in Chrome 61 branch.

Ok, will do.

alexeykuzmin:
Fail on Win:
> FAILED: obj/chromiumcontent/webkit.lib 
D:/build/vendor/depot_tools/python276_bin/python.exe ../../build/toolchain/win/tool_wrapper.py link-wrapper environment.x86 False lib.exe /nologo /ignore:4221 /OUT:obj/chromiumcontent/webkit.lib @obj/chromiumcontent/webkit.lib.rs
p
LINK : MSIL .netmodule or module compiled with /GL found; restarting link with /LTCG; add /LTCG to the link command line to improve linker performance
obj/chromiumcontent/webkit.lib : fatal error LNK1248: image size (10004143E) exceeds maximum allowable size (FFFFFFFF)

cifratila:
@alexeykuzmin I remember someone from the Summit (Google/Opera.TV) was saying that we will end up having this issue at linking, maybe we can follow up with Opera TV guys and see if they had a solution for this.

alexeykuzmin:
@cifratila I guess we can just split webkit lib into several libs.

alexeykuzmin:
@zcbenz 

> obj/chromiumcontent/webkitcoreremainings.lib : fatal error LNK1248: image size (100D429FE) exceeds maximum allowable size (FFFFFFFF)

alexeykuzmin:
@jkleinsc 
Both Windows build failed on uploading static library archive to Amazon S3.
Could it be caused by the size of the archive?

> [03:10:05] Copying D:\build\libchromiumcontent-static.zip to *****/libchromiumcontent/win/ia32/7205b3e096f591da55e1fd39725602b4feaccca4/libchromiumcontent-static.zip
<...>
[03:10:05] socket.error: [Errno 10054] An existing connection was forcibly closed by the remote host

zcbenz:
@alexeykuzmin I'm going to try using multipart upload.

zcbenz:
@jkleinsc I just noticed that you were also fixing the multipart problem of s3put on the chrome61 branch, I took a different approach here by removing the dependency of filechunkio module, which does not require using `pip` on the machines (we vendored the `boto` in libchromiumcontent to avoid this in the first place).

alexeykuzmin:
@jkleinsc If you're ok with the approach Cheng used here, let's merge it.

PR #365, opened 2017-10-01
Collect more shared libraries on Linux

alexeykuzmin:
Sonames do not always match '*.so' pattern,
sometimes they have a version number as a suffix, e.g. "libfreetype.so.6".


alexeykuzmin:
@zcbenz Can you please take a look?

alexeykuzmin:
@zcbenz Thanks! I'll merge then.

PR #371, opened 2017-10-10
Enable multipart upload to Amazon S3

alexeykuzmin:


alexeykuzmin:
@jkleinsc Shell we merge it? )

PR #374, opened 2017-10-13
Add patch for pdf support in sub frames

ahmedmohamedali:
This PR is a proposal to fix the issue https://github.com/electron/electron/issues/9192
The patch is fixing the following issues:

1) The  sub frame inherit the bindings from the main frame which does not have the web ui bindings.
Th fix is to reset the bindings fro the sub frame during the navigation

2) The user needs to set a name for the sub frame otherwise the web ui will be displayed in the main frame.  This limitation is removed by generating a name from the id for the sub frame when it hosts a web ui. A better approach would be to use the frame node id but this need a change in the chromium design for web ui which is based on the frame name.

3)  Messages are passed only to the web ui in the main frame. The fix is to iterate recursively the child frames and pass the message.

### DIFF
--- a/patches/042-webui_in_subframes.patch
+++ b/patches/042-webui_in_subframes.patch
@@ -79,7 +79,7 @@ index 4fa796b38b3f..376e8e143770 100644
 +	}
 +
 +	size_t child_count = parent_frame_host->frame_tree_node()->child_count();
-+	for (int i = 0; i < child_count; ++i) {
++	for (size_t i = 0; i < child_count; ++i) {
 +		RenderFrameHostImpl *sub_frame_host =
 +			parent_frame_host->frame_tree_node()->child_at(i)->current_frame_host();
 +		SenMessageToFrameTreeWebUIs(sub_frame_host, message, dispatch_count);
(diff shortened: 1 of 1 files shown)


ahmedmohamedali:
Hi @deepak1556 ,
Can you or someone from electron team start the review of this PR ? 
Thanks.

jkleinsc:
Closing and reopening this PR to trigger CI builds

ahmedmohamedali:
Hi electron team,

Can you please me let me know what’s the next step for this PR?

Thanks.

deepak1556:
Hi! @ahmedmohamedali sorry for the delayed response. The changes look good to me, I am going to verify the patch this weekend. Seems like the CI failed to generate builds to test locally. Thanks!

ahmedmohamedali:
Thanks @deepak1556  for approving this pull request.
Can you also approve the electron part #10793 ? Thanks.

zcbenz:
The failing tests are flaky, the build is actually fine. I'm merging this with them ignored.

zcbenz:
@alexeykuzmin A notice that this PR is adding a new patch.

alexeykuzmin:
@zcbenz Thanks!

PR #377, opened 2017-10-23
Add patch for nvidia GPU's on High Sierra

MarshallOfSound:
/cc @bpasero


jkleinsc:
@poiru if someone wants to patch it for 1.7, yes we can backport it there

poiru:
@jkleinsc Happy to do it! Seems like this was reverted though?

jkleinsc:
@poiru I was a bit aggressive in merging.  We need to test it against an Electron build before merging it in.  In talking to @MarshallOfSound and others today, this fix is an upstream quick/hacky fix from chromium canary that simply blacklists certain gpu features, so while it fixes the issue its a temporary fix at best.

alexeykuzmin:
>  if someone wants to patch it for 1.7, yes we can backport it there

https://github.com/electron/libchromiumcontent/pull/379
@jkleinsc, @poiru

PR #379, opened 2017-10-23
Add a patch to fix rendering issues with Nvidia GPU on High Sierra

alexeykuzmin:
https://crbug.com/773705

PR to the `master`: #382


alexeykuzmin:
@MarshallOfSound FYI
If your fix in the `master` turns out to be fine, then we should probably merge this too.

jkleinsc:
Corresponding Electron PR:
electron/electron#10898

PR #380, opened 2017-10-23
Temporarily turn off escape analysis in V8's Turbofan

alexeykuzmin:
Fixes a security vulnerability in V8's Turbofan engine.
For detailed description see
https://blogs.technet.microsoft.com/mmpc/2017/10/18/browser-security-beyond-sandboxing/

https://crbug.com/765433
https://chromium-review.googlesource.com/670563

### DIFF
--- a/patches/v8/turn_off_escape_analysis.patch
+++ b/patches/v8/turn_off_escape_analysis.patch
@@ -1,0 +1,13 @@
+diff --git a/src/flag-definitions.h b/src/flag-definitions.h
+index e6e9b7eb7a..a304326143 100644
+--- a/src/flag-definitions.h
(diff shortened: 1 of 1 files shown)


alexeykuzmin:
> Does this need to go into master as well?
@ckerr, sure it does. Here is a PR: #381.
I also wanted to create a PR to 1-8-x branch, but apparently we don't have one )

alexeykuzmin:
Turbofan is disabled by default in V8 < 5.9, so we don't need to do anything for Electron < 1.8.x.
https://v8project.blogspot.ru/2017/05/launching-ignition-and-turbofan.html

Closed.

PR #382, opened 2017-10-24
Add patch for nvidia GPU's on High Sierra (redo #377)

jkleinsc:
This is a redo of @MarshallOfSound's PR #377 that was merged into master a bit prematurely by myself.  This needs to be tested against a 1.8 Electron build before merging.


jkleinsc:
Electron tests pass in electron/electron#10923, except for osx-64 and I was able to manually verify those tests pass, so I am going to merge this in.

PR #387, opened 2017-11-08
Add freetype 2.8.1 patch

### LINKED ISSUE #384: Subpixel font rendering with freetype 2.8.1
With updated libfreetype fonts look "dirty" and painful to read due background artifacts. Here is a screenshot from unmodified `electron-quick-start`:

spectacle h23809

The issue affects all the Linux users with new version of the library. See: atom/atom#15737, Microsoft/vscode#35675.

It seems to be [upstream Skia issue](https://bugs.chromium.org/p/skia/issues/detail?id=6663). The fix could probably be backported from newer Chromium versions.

MarshallOfSound:
Fixes #384

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,46 @@
+pipeline {
+  agent none
+  stages {
+    stage('Build') {
+      parallel {
+        stage('libchromiumcontent-osx') {
+            agent {
+              label 'osx-libcc'
+            }
+            environment {
+              LIBCHROMIUMCONTENT_GIT_CACHE='/Volumes/LIBCC_CACHE'
+            }
+            steps {
+              sh 'script/cibuild'
+            }
+            post {
+              always {
+                archive 'libchromiumcontent.zip'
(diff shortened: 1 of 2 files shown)


MarshallOfSound:
@alexeykuzmin When this is merged it can be deleted in the chromium 61 branch (this patch is in chromium 61+)

alexeykuzmin:
@MarshallOfSound 👍

alexeykuzmin:
@MarshallOfSound Should it be merged? All the checks seem to be fine.

MarshallOfSound:
@alexeykuzmin Yeah, if someone approves all 3 PR's we can merge them in 👍

alexeykuzmin:
@ckerr If a patch just backports something from a newer version as is,
it will simply fail when we try to apply it to that version. It actually happens almost every Chromium and Node.js upgrade.
But if a patch is a temporary workaround, I would prefer to have an open issue to remove it when it's no longer needed.

PR #391, opened 2017-11-15
[DO NOT MERGE] Don't use custom libc++

alexeykuzmin:


alexeykuzmin:
We have to use libstdc++ newer than 4.8 [1] otherwise the build would fail on Linux.

[1] https://chromium-review.googlesource.com/c/chromium/src/+/583635

PR #394, opened 2017-11-21
Add helper script for backporting chromium patchsets

MarshallOfSound:
This makes is wayyyyy easier to backport chromium patchsets, syntax is like so:

```bash
python script/backport-patchset.py 2772283002 webrtc_crash_fix
```

This will pull down the patches for the given chromium code review ID and make a new patch in the patches directory for it 😄

### DIFF
--- a/script/backport-patchset.py
+++ b/script/backport-patchset.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python
+
+import os
+import re
+import sys
+import urllib2
+
+SOURCE_ROOT = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
+PATCHES_DIR = os.path.join(SOURCE_ROOT, 'patches')
+
+def main():
+  if len(sys.argv) < 3:
+    print("Usage: backport-patchset.py [code-review-id] [patch-name]")
+    return 1
+  codereview_url = "https://codereview.chromium.org/" + sys.argv[1]
+
+  codereview_response = urllib2.urlopen(codereview_url)
+  codereview = codereview_response.read()
+  diff_url_re = re.compile("/download/issue" + sys.argv[1] + "_[0-9]+\.diff", re.MULTILINE)
+  diff_url_result = diff_url_re.search(codereview)
+  if diff_url_result is None:
+    print("Failed to find diff for: " + sys.argv[1])
+    return 1
+  diff_url = "https://codereview.chromium.org" + diff_url_result.group(0)
+
+  diff_response = urllib2.urlopen(diff_url)
+  diff_content = diff_response.read()
+  diff_lines = diff_content.split("\n")[1:]
+  diffs = []
+  current_diff = []
+  for current_diff_line in diff_lines:
+    if re.match('Index: ', current_diff_line) is not None:
+      diffs.append(current_diff)
+      current_diff = []
+    else:
+      current_diff.append(current_diff_line)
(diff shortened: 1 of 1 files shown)


script/backport-patchset.py
```diff
@@ -0,0 +1,59 @@
...
+import re
+import sys
+import urllib2
+
+SOURCE_ROOT = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
+PATCHES_DIR = os.path.join(SOURCE_ROOT, 'patches')
+
+def main():
+  if len(sys.argv) < 3:
+    print("Usage: backport-patchset.py [code-review-id] [patch-name]")
```
ckerr: Since braces are for optional arguments, remove the braces here.

script/backport-patchset.py
```diff
@@ -0,0 +1,59 @@
...
+    return 1
+  codereview_url = "https://codereview.chromium.org/" + sys.argv[1]
```
ckerr: _(trivial)_ `sys.argv[1]` is referred to three times. It would be clearer to use a local instead, e.g.

`review_id = sys.argv[1]`

script/backport-patchset.py
```diff
@@ -0,0 +1,59 @@
...
+
+import os
+import re
+import sys
+import urllib2
+
+SOURCE_ROOT = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
+PATCHES_DIR = os.path.join(SOURCE_ROOT, 'patches')
+
+def main():
```
ckerr: The script overall LGTM and very handy.

It does, though, assume a lot of prior knowledge into how we do chromium backports. It would be nice for future contributors if `len(sys.argv) == 1` returned a brief help explaining how, when, and why this should be used.

script/backport-patchset.py
alexeykuzmin: Maybe [argparse](https://docs.python.org/2.7/library/argparse.html) could be useful here?

script/backport-patchset.py
```diff
@@ -0,0 +1,59 @@
...
+  if len(sys.argv) < 3:
+    print("Usage: backport-patchset.py [code-review-id] [patch-name]")
+    return 1
+  codereview_url = "https://codereview.chromium.org/" + sys.argv[1]
+
+  codereview_response = urllib2.urlopen(codereview_url)
```
alexeykuzmin: `main()` is really huge now, maybe we could extract getting a diff by a `code-review-id` to a function?

script/backport-patchset.py
```diff
@@ -0,0 +1,59 @@
...
+      current_diff = []
+    else:
+      current_diff.append(current_diff_line)
+  diffs.append(current_diff)
+  for i in range(len(diffs)):
+    diffs[i] = '\n'.join(diffs[i])
+
+  patch_file = os.path.join(PATCHES_DIR, get_next_patch_number(PATCHES_DIR) + "-" + sys.argv[2] + ".patch")
+  with open(patch_file, "w") as file:
+    file.write("# Backported patch from Chromium Patchset: " + codereview_url + "\n")
```
alexeykuzmin: Can this comment break anything? I don't remember seeing any comment in patch files in the libcc.

alexeykuzmin:
@MarshallOfSound Can we close this PR? ) Since #512 has merged.

PR #396, opened 2017-11-22
Collect objects from third_party/libdrm

zcbenz:
The `third_party/libdrm` is build and used by some code of content module, but the linker could build electron without it since electron is not using related code. However the linker of mips64el is not so advanced to optimize the dependency out and we have to ship it as part of libcc.


zcbenz:
I have tested this build with Electron and both Debug and Release builds were working.

PR #397, opened 2017-11-23
Add patch to fix the pagesize error in v8 on MIPS64EL

zcbenz:
See https://github.com/electron/electron/pull/11208#issuecomment-346545546.


ClarkWang-2013:
@zcbenz I have a simple modify, and only considered loongson platform.if you merge to libchromiumcontent, you must consider other platform(such as, x86 is 4K pagesize). 
I think you set 16384 directly is not acceptable.

zcbenz:
@ClarkWang-2013 This patch is only applied when building for mips64el, so it is fine.

zcbenz:
I'm merging this since it solves the v8 crash and does not affect other builds.

PR #403, opened 2017-11-27
Reduce number of disabled DCHECKs

alespergl:
This limits the scope of disabled DCHECKs to just a handful of (still fairly sizable) parts of libchromiumcontent, where Electron currently experiences DCHECK failures.

I'm introducing a build-time flag `ELECTRON_NO_DCHECK` which, when defined, disables DCHECK.

The following GN targets have the flag applied:
- `base:base`
- `content/browser:browser`
- `content/child:child`
- `device/geolocation:geolocation`
- `net:net`
- `third_party/WebKit/Source/core/loader:loader`
- `third_party/WebKit/Source/platform:platform`
- `ui/base:base`
- `url:url`

Additionally these files have debug checks explicitly commented out:

- `content/browser/frame_host/render_frame_host_impl.cc`
- `ppapi/host/ppapi_host.cc`
- `storage/browser/quota/client_usage_tracker.cc`
- `third_party/WebKit/Source/platform/wtf/ThreadRestrictionVerifier.h`
- `ui/latency/latency_histogram_macros.h`

P.S.: I'm also sneaking in removal of one obsolete patch file.

PR to Electron: electron/electron#11271


alexeykuzmin:
"Removed obsolete `022-libgtkui.patch`" seems to be unrelated to DCHECK, maybe it would make sense to remove it in a separate pr?

alexeykuzmin:
Probably we should try to build Electron and run its tests using libcc built from this branch.

alexeykuzmin:
> The following GN targets have the flag applied:
>
> `base:base`

it kinda scares me that we have the very basic stuff broken =/

ckerr:
> it kinda scares me that we have the very basic stuff broken =/

It was broken before, too. This patch is part of what we discussed last week about nibbling away at this, enabling and fixing the DCHECKS in smaller chunks

alexeykuzmin:
@ckerr sure
I was trying to say that before we knew that _some_ stuff is broken, but now we know that _the very base_ stuff is broken.

alexeykuzmin:
Are we waiting for other approvals or can we merge now?
I really want this change in the Ch62 branch.

alespergl:
Let us first run the test suite, in case there is more stuff to disable with this.

deepak1556:
I have created https://github.com/electron/electron/pull/11271 to start working with this, I think we might want to disable dchecks in v8. Current fails on dcheck in `url` target.

alespergl:
I quickly checked the failure
`[6975:1127/210918.696407:FATAL:url_util.cc(490)] Check failed: !scheme_registries_locked. Trying to add a scheme after the lists have been locked.`

I will disable DCHECK for the relevant GN target, but the problem is in how [`protocol.registerStandardSchemes`](https://electronjs.org/docs/api/protocol#protocolregisterstandardschemesschemes-options) is implemented. It will need to be refactored or removed to fix this properly. Schemes need to be added in [`AtomContentClient::AddAdditionalSchemes`](https://github.com/electron/electron/blob/124667c29b249b5f79c57c8abbf2912157370cc7/atom/app/atom_content_client.cc#L201) and nowhere else.

alexeykuzmin:
> I will disable DCHECK for the relevant GN target, but the problem is in how protocol.registerStandardSchemes is implemented. It will need to be refactored or removed to fix this properly. Schemes need to be added in AtomContentClient::AddAdditionalSchemes and nowhere else.

@alespergl Can you create a task to fix it please?

alespergl:
https://github.com/electron/electron/issues/11273

alexeykuzmin:
btw, maybe it would make sense to rebase this branch onto the latest master
to get Jumbo build enabled, it could speed up the builds.

alespergl:
The reported failure was [this one in `ThreadRestrictionVerifier.h`](https://cs.chromium.org/chromium/src/third_party/WebKit/Source/platform/wtf/ThreadRestrictionVerifier.h?l=94). That's why I disabled it. The DCHECKs you are referring to don't seem to be failing (so far). Let's see if we can get to a state where there are no DCHECK failures in tests and then work backwards from there.

alexeykuzmin:
@alespergl 

```diff
-  DCHECK_NE(-1, idr) << " path: " << path;
+  // DCHECK_NE(-1, idr) << " path: " << path;
```

I don't think that we can simply ignore it =/
"chrome://resources/..." is Chromium internal "protocol" to access shared resources like libraries, fonts, images, etc. from internal pages. If we get a request to load some Polymer stuff (used on most internal pages of Chromium), it means we somehow load very unexpected pages.
I created an issue for Electron electron/electron#11289.

alespergl:
We can't ignore any of these. Nevertheless, we have been doing that so far. :)
Thanks for creating that issue, at least this one will hopefully be fixed soon.

alexeykuzmin:
We should probably update libcc reference in the electron/electron#11271 and merge it.

alespergl:
No worries, I'll do that when the merged commit is built. ❄️

PR #404, opened 2017-11-30
Split builds into static and shared builds

jkleinsc:
As requested by @alexeykuzmin, this PR  breaks up builds of shared_library and static_library configurations into separate builds which will allow us to speed up our builds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,10 +1,46 @@
 version: 2.0
 jobs:
-  libchromiumcontent-linux-x64:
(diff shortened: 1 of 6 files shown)


alexeykuzmin:
I think that someday we should change the names of the libcc configurations
to something like "debug" and "release" or similar.
Names "static" and "shared" are quite technical and might be not clear for newcomers.
@zcbenz what do you think?

zcbenz:
> I think that someday we should change the names of the libcc configurations
> to something like "debug" and "release" or similar.

👍

PR #407, opened 2017-12-04
Upgrade to Chromium 63

alexeykuzmin:
**63.0.3239.150**

Electron PR: https://github.com/electron/electron/pull/11459
Project board: https://github.com/electron/electron/projects/7


patches/v8/007-export_platform.patch
```diff
@@ -0,0 +1,25 @@
+diff --git a/include/v8-platform.h b/include/v8-platform.h
+index 74630b6f19..9704c35146 100644
+--- a/include/v8-platform.h
++++ b/include/v8-platform.h
+@@ -10,6 +10,8 @@
+ #include <memory>
+ #include <string>
+ 
++#include "v8.h"
```
deepak1556: Can we have the `V8_EXPORT` declarations moved out to a separate file `v8-export.h` and include it in `v8.h` and `v8-platform.h` instead ?

alexeykuzmin: @deepak1556 I believe that patches should be as small as possible to minimise the maintenance cost.

> `V8_EXPORT` declarations moved out to a separate file `v8-export.h`

It probably make sense, but it should be done the V8 repo.

Issue #360, opened 2017-09-24
Run Chromium tests


alexeykuzmin:
Big stuff. The plan is:

- [ ] Identify all test "suites" that cover the code we use from Chromium (i.e. content_unittests, etc.)
- [ ] List failing tests from all "suites"
- [ ] Fix failing tests (either modify related code in a patch or modify/disable the test)
- [ ] _optional_ Add tests for patches we have
- [ ] Build and run tests on CI bots

Chromium tests documentation:
- [Running tests locally](https://www.chromium.org/developers/testing/running-tests)
- [Layout Tests](https://chromium.googlesource.com/chromium/src/+/master/docs/testing/layout_tests.md#running-the-tests)

Related PRs: #440, #441, #443

/cc @jkleinsc, @tonyganch, @zcbenz, @groundwater, @gavignus

alexeykuzmin:
We can start with the `content_unittests` and go directly to the step five: "Build and run tests on CI bots." It means we need to 1) build `content_unittests` target, 2) run it, 3) parse results, 4) display results in a human-readable form.

@jkleinsc I would need your help to actually run the tests on CI machines.
Not right now, but in the nearest future.

alexeykuzmin:
Build config for tests:

```
is_debug = false
is_component_build = true
dcheck_always_on = true
symbol_level = 1
compute_grit_inputs_for_analyze = true
enable_update_channel = false
enable_nacl = false
```

alexeykuzmin:
### Failing tests

#### content_unittests
- DOMStorageMapTest.EnforcesQuota

jkleinsc:
@alexeykuzmin is there any thought about running Electron test suite as well?  In theory we can build Electron from the compiled libcc and run the Electron test suite.  I think maybe all the parts are there because on the Electron build you can specify where libcc is.

alexeykuzmin:
@jkleinsc so you're suggestion on every change in libcc run
build libcc → run libcc tests → build Electron (with this libcc) → run Electron tests ?
Makes a lot of sense.

It would require some non-trivial changes in out CI setup though. But will be extremely useful.

jkleinsc:
Yes, that way we verify libcc PR is ok to go without having to have electron PR before we can verify it.

alexeykuzmin:
Ok. So, on every commit in the _libcc_ repo we should trigger _not libcc_, but _Electron_ build.
Electron builds shouldn't expect libcc binaries to be always available for download, but instead
[...]

PR #411, opened 2017-12-06
[RFC] [WIP] Test Electron after building

jkleinsc:
In #360, @alexeykuzmin and I were talking about running Electron tests from a libcc build.  This PR does that, but not everything about it is ideal.  Both CircleCI and Jenkins pipelines allow us to create workflows that wait until both the static and shared builds are done, but for AppVeyor we can't really do that.  So for AppVeyor we build a debug Electron build from master and run tests against that at the end of the shared builds.  CircleCI and Jenkins build a release Electron and run tests against that.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,6 +70,52 @@ jobs:
       - store_artifacts:
           path: libchromiumcontent-static.zip
 
+  libchromiumcontent-linux-x64-test:
+    docker:
+      - image: electronbuilds/electron:0.0.3
+        environment:
+          TARGET_ARCH: x64
+          DISPLAY: ':99.0'
+    resource_class: 2xlarge
+    steps:
+      - run:
+          name: Checkout Electron for testing
+          command: |
+            git clone --depth 1 https://github.com/electron/electron.git
+      - run:
+          name: Setup for headless testing
+          command: sh -e /etc/init.d/xvfb start
+      - run:
+         name: Bootstrap Electron
+         command: |
+            cd electron
+            echo 'export LIBCHROMIUMCONTENT_COMMIT=$CIRCLE_SHA1' >> $BASH_ENV
+            echo "Bootstrapping Electron for release build with commit $CIRCLE_SHA1"
+            script/bootstrap.py --target_arch=$TARGET_ARCH
+      - run:
+         name: Build Electron release
+         command: |
+            cd electron
+            script/build.py -c R
+      - run:
+          name: Test Electron
+          environment:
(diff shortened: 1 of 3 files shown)


alexeykuzmin:
Right now we run tests only against Electron debug builds (which use libcc shared_library build).
This is good because debug builds would fail on dchecks, but on the other side they're quite slow and it might be the reason of timeouts.

So I would prefer to have a different (a third, yeah...) libcc configuration for tests: release build + enabled dchecks. It won't be useful for anything except tests, but resulting binary have better performance than a debug build but still will fail on dchecks. And that's how Chromium builds its test binaries.

@jkleinsc what do you think?

alexeykuzmin:
>  (a third, yeah...)

It won't increase the number of libcc builds, and CI machines load though.
There will be no need in building release builds for every commit/PR/branch/whatever, no one would normally need it.

jkleinsc:
@alexeykuzmin I'm good with adding a 3rd config.  Splitting out the shared and static builds has sped up things and adding a 3 config would be ok by me.   What if PRs built just this new config and then on branches we care about (eg master, electron-1-8-x) we build all 3?

alexeykuzmin:
> What if PRs built just this new config and then on branches we care about (eg master, electron-1-8-x) we build all 3?

_let's call this new config "testing"_

Testing builds are good for running tests, but if something goes wrong, you would probably want to have debug (aka shared_library) build to understand and fix the problem.
If a debug build isn't available for download you would have to build libcc locally that takes a lot of time.

> on branches we care about (eg master, electron-1-8-x) we build all 3

Building debug and testing builds for them would make total sense.
For release builds I'm not sure, maybe we can build them on demand only.

jkleinsc:
> Building debug and testing builds for them would make total sense.
I'm not sure about release builds, maybe we can build them on demand only.

So always build `testing` and `debug` builds and make `release` builds an on demand build.  Yeah, we could do that.  I do wonder though if it would just make sense to run all three builds on those special branches (master, electron-1-8-x) because we then would always have a release build available when a PR is merged in.

jkleinsc:
@alexeykuzmin as far as testing libcc with Electron with the `testing` build, would the Electron build be a release build or a debug build?

alexeykuzmin:
> as far as testing libcc with Electron with the `testing` build, would the Electron build be a release build or a debug build?

release
`testing` libcc will be almost identical to the current `static_library`

alexeykuzmin:
on the second thought... we will probably have to create another Electron config to match that [release build + dchecks] logic. I'm sure if it is possible with gyp, will have to check =/

PR #413, opened 2017-12-11
Enable additional DCHECKs

deepak1556:
Get some libcc builds to work on https://github.com/electron/electron/pull/11396


patches/002-dcheck.patch
```diff
@@ -36,32 +36,17 @@ These files have debug checks explicitly commented out:
   ui/latency/latency_histogram_macros.h
 
 
-diff --git a/base/BUILD.gn b/base/BUILD.gn
```
alespergl: Before merging, please update the list in the above patch description by removing `base:base` from it.

patches/002-dcheck.patch
```diff
@@ -36,32 +36,17 @@ These files have debug checks explicitly commented out:
...
- }
- 
- buildflag_header("debugging_flags") {
 diff --git a/base/logging.h b/base/logging.h
 index 582fb89..24d3e82 100644
 --- a/base/logging.h
 +++ b/base/logging.h
 @@ -850,7 +850,7 @@ const LogSeverity LOG_DCHECK = LOG_INFO;
- 
+
```
alespergl: Don't remove this trailing whitespace. It's part of the diff format.

deepak1556:
@zcbenz can we move the `clipboard_win.cc` patch to `002-dcheck.patch` ? It would be better to have them in one file.

PR #417, opened 2017-12-15
[webrtc][mac] cherry-pick fix for laggy mouse during desktop capture

CapOM:
The mouse is being capture twice, one time with a custom way
and one time with a built-in Mac API. The later was not known
so the mouse was wrongly captured twice.
Also the later way causes the cursor to move slowly on the
machine that is being captured.

The solution is just to disable this build-in way to just
keep the former way.

https://webrtc-review.googlesource.com/c/src/+/30902

### DIFF
--- a/patches/third_party/webrtc/001-fix_laggy_mouse_during_desktop_capture.patch
+++ b/patches/third_party/webrtc/001-fix_laggy_mouse_during_desktop_capture.patch
@@ -1,0 +1,27 @@
+diff --git a/modules/desktop_capture/screen_capturer_mac.mm b/modules/desktop_capture/screen_capturer_mac.mm
+index f48b30774..0fcdfc61f 100644
+--- a/modules/desktop_capture/screen_capturer_mac.mm
++++ b/modules/desktop_capture/screen_capturer_mac.mm
+@@ -966,8 +966,21 @@ void ScreenRefresh(CGRectCount count,
+             ScreenRefresh(count, rects, display_origin);
+           }
+         };
++
++    CFDictionaryRef properties_dict = CFDictionaryCreate(kCFAllocatorDefault,
++                                                         (const void* []){kCGDisplayStreamShowCursor},
++                                                         (const void* []){kCFBooleanFalse},
++                                                         1,
++                                                         &kCFTypeDictionaryKeyCallBacks,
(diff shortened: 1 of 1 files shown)


CapOM:
>> Could you please create another PR to master?
Done.

alexeykuzmin:
@CapOM Is there an issue for this in the [electron/electron tracker](https://github.com/electron/electron/issues)?

alexeykuzmin:
The patch is somewhat different from the changes in the webrtc CL https://webrtc-review.googlesource.com/c/src/+/30902, but I hope you know what you're doing )
CI build on Mac is green, so approved.

alexeykuzmin:
Red Windows build is irrelevant since the patch touches only Mac-specific code.

alexeykuzmin:
@CapOM If we decide to include it in the upcoming 1.8.x release, we'll create a cherry-pick to `electron-1-8-x` ourselves.
Thank you!

CapOM:
The difference is because webrtc/sdk/objc/Framework/Classes/Common/scoped_cftyperef.h does not exist but it is just used to wrap the dictionary into a ScopedCFTypeRef. But calling CFRelease manually is equivalent here.

I really hope it will be decided to cherry-pick the patch to electron-1-8-x

ckerr:
This sounds like a good candidate for 1-8-x but I don't feel confident in taking point on a Mac code review and and the "I hope you know what you're doing" comment above makes me want another reviewer. :confused:

@zcbenz could you take a look?

alexeykuzmin:
@ckerr I approved this patch because it's just a cherry-pick from the WebRTC repo, adopted to our old codebase, and made by the original author himself. The changes were reviewed and approved before they were merged into WebRTC's _master_ here https://webrtc-review.googlesource.com/c/src/+/30902.

zcbenz:
Yeah for cherry-picks from upstream I would only check whether the code is the same and whether it compiles, there is usually no need to review the code itself since it has been reviewed by the upstream.

ckerr:
@alexeykuzmin it was the 'this differs, I hope you know what you're doing' comment that gave me pause, but maybe you were making a joke and I misread the intent.

Either way, if you and @zcbenz are both happy that's good enough for me

PR #428, opened 2018-01-07
Backport a WebRTC patch for video cropping to prevent freezes.

ajmacd:
Fixes the problem documented in this bug:
https://bugs.chromium.org/p/webrtc/issues/detail?id=6958

where the encoder receives frame sizes it's not expecting and errors
out. The fix landed in M61:
https://codereview.webrtc.org/2936393002

This applies the critical pieces of that change to crop and scale poorly
sized frames before they hit the encoder.

Verified the fix with Slack video calling. It's easy to reproduce by
requesting a 641x480 getUserMedia stream from the camera. The logs will
be filled with this error:
```
Incoming frame doesn't match set resolution. Dropping.
```

### DIFF
--- a/patches/third_party/webrtc/002-crop_video_to_prevent_freezes.patch
+++ b/patches/third_party/webrtc/002-crop_video_to_prevent_freezes.patch
@@ -1,0 +1,75 @@
+diff --git a/video/vie_encoder.cc b/video/vie_encoder.cc
+index 3a848f852..825451084 100644
+--- a/video/vie_encoder.cc
++++ b/video/vie_encoder.cc
+@@ -494,6 +494,20 @@ void ViEEncoder::ReconfigureEncoder() {
+       encoder_config_.video_stream_factory->CreateEncoderStreams(
+           last_frame_info_->width, last_frame_info_->height, encoder_config_);
+ 
++  // TODO(ilnik): If configured resolution is significantly less than provided,
++  // e.g. because there are not enough SSRCs for all simulcast streams,
++  // signal new resolutions via SinkWants to video source.
++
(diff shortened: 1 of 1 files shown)


patches/third_party/webrtc/002-crop_video_to_prevent_freezes.patch
```diff
@@ -0,0 +1,75 @@
...
++  if (crop_width_ > 0 || crop_height_ > 0) {
++    int cropped_width = video_frame.width() - crop_width_;
++    int cropped_height = video_frame.height() - crop_height_;
++    rtc::scoped_refptr<I420Buffer> cropped_buffer =
++        I420Buffer::Create(cropped_width, cropped_height);
++    // TODO(ilnik): Remove scaling if cropping is too big, as it should never
++    // happen after SinkWants signaled correctly from ReconfigureEncoder.
++    if (crop_width_ < 4 && crop_height_ < 4) {
++      cropped_buffer->CropAndScaleFrom(
++          *video_frame.video_frame_buffer(), crop_width_ / 2, crop_height_ / 2,
```
ajmacd: This and line 47 are slightly modified to remove ->ToI420() from the original change, because the I420Buffer interface changed in the interim.

alexeykuzmin:
@ckerr It might make sense to have it in the upcoming 1.8.x release.

alexeykuzmin:
@zcbenz, can we ignore the mips64el build failure?

ckerr:
mips64el build failure question answered @ https://github.com/electron/libchromiumcontent/pull/433#issuecomment-358872830, so merging this one

PR #430, opened 2018-01-08
Revert #349 for master branch

zcbenz:
#349 made some accelerators like `Ctrl+Shift+A` and `Ctrl+Shift+F11` show without the `Shift` key. I tried tweak the patch but can not find a easy fix, so I'm reverting it since it is preventing 1.8.x stable.

Refs https://github.com/electron/electron/issues/11429.


alexeykuzmin:
cc @tonyganch

PR #432, opened 2018-01-08
Backport WebRTC fix for connection failures with restrictive firewalls (master).

ajmacd:
Addresses the problem described here:
https://bugs.chromium.org/p/webrtc/issues/detail?id=8102

where, after DNS resolution fails with a restrictive firewall, TLS
connections fail immediately.

As pointed out here in the bug:
https://bugs.chromium.org/p/webrtc/issues/detail?id=8102#c7

the problem is fixed with this change:
https://chromium-review.googlesource.com/c/external/webrtc/+/611520/9/webrtc/p2p/base/turnport.cc

which allows the connection attempt to proceed using the hostname as intended.


ajmacd:
Also PR'd against `1-8-x` here:
https://github.com/electron/libchromiumcontent/pull/433

PR #433, opened 2018-01-08
Backport WebRTC fix for connection failures with restrictive firewalls (1-8-x).

ajmacd:
Addresses the problem described here:
https://bugs.chromium.org/p/webrtc/issues/detail?id=8102

where, after DNS resolution fails with a restrictive firewall, TLS
connections fail immediately.

As pointed out here in the bug:
https://bugs.chromium.org/p/webrtc/issues/detail?id=8102#c7

the problem is fixed with this change:
https://chromium-review.googlesource.com/c/external/webrtc/+/611520/9/webrtc/p2p/base/turnport.cc

which allows the connection attempt to proceed using the hostname as intended.

### DIFF
--- a/patches/third_party/webrtc/003-connect_using_hostname_on_dns_failure_over_tls.patch
+++ b/patches/third_party/webrtc/003-connect_using_hostname_on_dns_failure_over_tls.patch
@@ -1,0 +1,24 @@
+diff --git a/p2p/base/turnport.cc b/p2p/base/turnport.cc
+index 901516922..52280c4a3 100644
+--- a/p2p/base/turnport.cc
++++ b/p2p/base/turnport.cc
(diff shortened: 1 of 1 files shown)


ajmacd:
Right, I should have pointed out why I didn't add that one: it's present in M61 but not M59, so I excluded it in the interests of using the same patch. (And since there is no production impact of course.)

alexeykuzmin:
@ckerr It might make sense to have it in the upcoming 1.8.x release.

alexeykuzmin:
@zcbenz, can we ignore mips64el build failure?

zcbenz:
Yeah please ignore the mips64el failures, I think I have already disabled the mips64el on electron-1-8-x branch.

PR #435, opened 2018-01-17
Fix hanged large parallel POSTS in H2

alexeykuzmin:
This CL makes sure that streams that get stalled on send are being put
back in the queue if they cannot send when attempting to unstall them.
https://chromium-review.googlesource.com/649849

Fixes electron/electron#11616.

### DIFF
--- a/patches/075-cherrypick_f1b7017a.patch
+++ b/patches/075-cherrypick_f1b7017a.patch
@@ -1,0 +1,92 @@
+diff --git a/net/spdy/chromium/spdy_session.cc b/net/spdy/chromium/spdy_session.cc
+index 665cc54e7fa0..e8271d9c1428 100644
+--- a/net/spdy/chromium/spdy_session.cc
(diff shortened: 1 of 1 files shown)


patches/075-cherrypick_f1b7017a.patch
```diff
@@ -0,0 +1,93 @@
+96b36db1d2dfb5c4c69679a8a7e60e158a770668
```
deepak1556: I guess this is not related to the patch ?

alexeykuzmin: I have no idea what it is and how it got here ) Thanks!

alexeykuzmin: Removed.

alexeykuzmin:
I'm going to merge this without any other approvals because the changes have already been reviewed before being merged to Chromium and the patch matches the CL exactly.

PR #436, opened 2018-01-17
[1.8.x] Fix hanged large parallel POSTS in H2

alexeykuzmin:
This CL makes sure that streams that get stalled on send are being put
back in the queue if they cannot send when attempting to unstall them.
https://chromium-review.googlesource.com/649849

Fixes electron/electron#11616.

Fixed in libcc/master in #435.


alexeykuzmin:
cc @ckerr

PR #439, opened 2018-01-24
Turn off use_nss_certs for mips64el

zcbenz:
Linking with nss was causing crashes on mips64el platforms, the crash was coming from the system libsqlite3 which seemed to conflict with the forked libsqlite3 version inside Chromium:

```
(gdb) bt
#0  0x0000000000000000 in  ()
#1  0x000000ffe84e5bfc in sqlite3Malloc () at /lib64/libsqlite3.so.0
#2  0x000000ffe85d2cac in openDatabase () at /lib64/libsqlite3.so.0
#3  0x000000ffea8a8c84 in sdb_init () at /lib64/libsoftokn3.so
```

Turning off `use_nss_certs` can temporarily solve this problem, with the downside of making Electron's certificate related APIs not work.

We can probably solve this cleanly once Chromium [supports building with system sqlite](https://bugs.chromium.org/p/chromium/issues/detail?id=22208), but before that we are going to use this workaround while still exploring other possible solutions.

This patch is only applied for mips64el, other targets are not affected at all.

This PR also backports the commit to split out static and shared builds, so compilation would be faster.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,10 +1,46 @@
 version: 2.0
 jobs:
-  libchromiumcontent-linux-x64:
+  libchromiumcontent-linux-x64-shared:
     docker:
       - image: electronbuilds/libchromiumcontent:0.0.4
         environment:
           TARGET_ARCH: x64
+          COMPONENT: shared_library
+    resource_class: 2xlarge
+    steps:
+      - checkout
+      - run:
+          name: Bootstrap
+          command: script/bootstrap
+      - run:
+          name: Update
+          command: script/update --clean -t $TARGET_ARCH
+      - run:
+          name: Build shared library
+          command: script/build -t $TARGET_ARCH -c $COMPONENT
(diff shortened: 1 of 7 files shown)


deepak1556:
There is also work on moving from NSS to BoringSSL for client cert verification in linux, should help solve this issue.

* https://bugs.chromium.org/p/chromium/issues/detail?id=338887
* https://bugs.chromium.org/p/chromium/issues/detail?id=604728
* https://bugs.chromium.org/p/chromium/issues/detail?id=394131

deepak1556:
Chromium doesn't have plans to support using system sqlite as per https://bugs.chromium.org/p/chromium/issues/detail?id=22208#c51 , we have to move away from NSS when possible to fix this.

zcbenz:
Chromium seems to have fixed the root cause of the crash:
https://bugs.chromium.org/p/chromium/issues/detail?id=807487

PR #440, opened 2018-01-25
Light refactoring of the libcc scripts

alexeykuzmin:


alexeykuzmin:
@zcbenz can you please take a look?

PR #442, opened 2018-01-29
WIP: Set _GLIBCXX_USE_CXX11_ABI=0 on Linux

zcbenz:
According to:
https://gcc.gnu.org/onlinedocs/libstdc++/manual/backwards.html
https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html

By defining `_GLIBCXX_USE_CXX11_ABI=0` we should be able to link with new versions of libstdc++ while keeping backward compatibility with the old ones.


zcbenz:
Got some weird crashes, it seems that it is mostly useful for keeping compatibility with old binaries, instead of generating compatible binaries.

PR #443, opened 2018-01-29
Enable unittests and browsertests

deepak1556:
This PR tries to accomplish the first three action items in https://github.com/electron/libchromiumcontent/issues/360

```
$ ./script/run_tests

40 tests run.
The following 7 tests failed:
gpu_unittests
blink_platform_unittests
base_unittests
content_unittests
net_unittests
compositor_unittests
content_browsertests
```

- [ ] Build test target on all platforms
  - [x] MacOS
  - [ ] Linux
  - [ ] Windows
- [ ] Fix failing tests
  - [ ] gpu_unittests
  - [ ] blink_platform_unittests
  - [ ] base_unittests
  - [ ] content_unittests
  - [ ] net_unittests
  - [ ] compositor_unittests
  - [ ] content_browsertests

/cc @alexeykuzmin

### DIFF
--- a/chromiumcontent/BUILD.gn
+++ b/chromiumcontent/BUILD.gn
@@ -351,4 +351,62 @@ if (is_electron_build && !is_component_build) {
     deps = [ ":targets" ]
   }
 
+  group("chromiumcontent_tests") {
+    testonly = true
+
+    deps = [
+      "//base:base_unittests",
+      "//cc:cc_unittests",
+      "//cc/blink:cc_blink_unittests",
+      "//content/test:content_browsertests",
+      "//content/test:content_unittests",
+      "//crypto:crypto_unittests",
+      "//device:device_unittests",
+      "//gin:gin_unittests",
+      "//gpu:gpu_unittests",
+      "//gpu/ipc/service:gpu_ipc_service_unittests",
+      "//ipc:ipc_tests",
+      "//media:media_unittests",
+      "//media/midi:midi_unittests",
+      "//media/mojo:media_mojo_unittests",
+      "//media/capture:capture_unittests",
+      "//mojo/common:mojo_common_unittests",
+      "//mojo/edk/system:mojo_system_unittests",
+      "//mojo/edk/test:mojo_public_bindings_unittests",
+      "//mojo/edk/test:mojo_public_system_unittests",
+      "//net:net_unittests",
+      "//ppapi:ppapi_unittests",
+      "//printing:printing_unittests",
+      "//skia:skia_unittests",
+      "//storage:storage_unittests",
+      "//sql:sql_unittests",
+      "//third_party/angle/src/tests:angle_unittests",
+      "//third_party/leveldatabase:env_chromium_unittests",
+      "//third_party/WebKit/public:test_support",
+      "//third_party/WebKit/public:all_blink",
+      "//tools/gn:gn_unittests",
+      "//ui/base:ui_base_unittests",
+      "//ui/compositor:compositor_unittests",
+      "//ui/display:display_unittests",
+      "//ui/events:events_unittests",
+      "//ui/gl:gl_unittests",
+      "//url:url_unittests",
+      "//url/ipc:url_ipc_unittests",
+      "//v8/test/unittests:unittests",
+    ]
+
+    if (is_linux) {
+      deps += [
+        "//net:disk_cache_memory_test",
+        "//sandbox/linux:sandbox_linux_unittests",
+      ]
+
+      if (use_dbus) {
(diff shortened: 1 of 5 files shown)


alexeykuzmin:
@deepak1556 

> Fix failing tests

We can probably simply disable them, merge, and fix the tests later.

alexeykuzmin:
@jkleinsc We will need your help to understand that should be done to display a list of all failed tests on all CI platforms, and fail builds if there're any.

chromiumcontent/BUILD.gn
```diff
@@ -360,6 +360,64 @@ if (is_electron_build && !is_component_build) {
     deps = [ ":targets" ]
   }
 
+  group("chromiumcontent_tests") {
```
alexeykuzmin: I think it would make sense to create separate groups unittests from browsertests.
Unittests are fast and it's easy to run them locally, browsertests are not that fast.

script/build
```diff
@@ -60,13 +60,15 @@ def main():
...
         continue
       out_dir = get_output_dir(SOURCE_ROOT, target_arch, component)
-      target = 'chromiumcontent:chromiumcontent'
-      subprocess.check_call([NINJA, '-C', os.path.relpath(out_dir), target], env=env)
+      subenv = env.copy()
       if component == 'static_library':
-        subenv = env.copy()
         subenv['CHROMIUMCONTENT_2ND_PASS'] = '1'
         target = 'chromiumcontent:libs'
-        subprocess.check_call([NINJA, '-C', os.path.relpath(out_dir), target], env=subenv)
```
alexeykuzmin: `static_library` actually should be built twice =(

script/build
```diff
@@ -60,13 +60,15 @@ def main():
...
+      elif component == 'tests':
+        target = 'chromiumcontent:chromiumcontent_tests'
+      else:
+        target = 'chromiumcontent:chromiumcontent'
+      subprocess.check_call([NINJA, '-C', os.path.relpath(out_dir), target], env=subenv)
```
alexeykuzmin: We probably need to make `target` a build script argument to allow building and running tests
in a different build configuration, e.g. `shared_library`.

I can do it in a separate PR.

alexeykuzmin:
@deepak1556 I just rebased it onto the latest master.

jkleinsc:
@alexeykuzmin if run_tests exits with an exit code other than 0 then all our CI platforms will mark the build as failed, so as long as run_tests is returning and exit code other that 0 on failure we should be good to go.

jkleinsc:
@deepak1556 @alexeykuzmin I was just looking at this to see what we need to change in our CI configs to support this and I wanted to verify something:  do the tests need the static build to run or will they/should they also run against the debug build?

alexeykuzmin:
> do the tests need the static build to run or will they/should they also run against the debug build

I want both options to be available for developers, but run on release-ish build on CI.
By "release-ish" I mean the thing we call "static_library" with DCHECKs enabled.

alexeykuzmin:
> @alexeykuzmin if run_tests exits with an exit code other than 0 then all our CI platforms will mark the build as failed, so as long as run_tests is returning and exit code other that 0 on failure we should be good to go.

Yeah, but I also want a list of failed tests to be displayed in a nice form on a CI UI, not only a raw output.

alexeykuzmin:
> By "release-ish" I mean the thing we call "static_library" with DCHECKs enabled.

So it's going to be a different build configuration.
I think it would be better to discuss "perfect" setup with more people, somewhere in Slack.

nitsakh:
`gpu_unittests` has only 1 failure on macos, namely, `GPUTestConfigTest.LoadCurrentConfig`. This test uses `GetCurrentOS` function from `gpu_test_config.cc` which supports only till macos Sierra. So this test should work on other platforms.

jkleinsc:
> Yeah, but I also want a list of failed tests to be displayed in a nice form on a CI UI, not only a raw output.

@alexeykuzmin we might be able to convert the test results into junit results which our CI platforms place fairly nicely with (though display varies per CI).  It looks like the test results are in JSON: https://chromium.googlesource.com/chromium/src/+/master/docs/testing/json_test_results_format.md and I'm wondering if we can use https://github.com/aaronbriel/jsonjunit to convert.

alexeykuzmin:
@jkleinsc with `--gtest_output=xml` flag test results will be saved in JUnit XML format.

```
  --gtest_output=xml[:DIRECTORY_PATH/|:FILE_PATH]
      Generate an XML report in the given directory or with the given file
      name. FILE_PATH defaults to test_detail.xml.
```

PR #445, opened 2018-02-05
Switch to `tar.bz2` for distribution package

alespergl:
The distribution packages are getting ever bigger. Especially after we switch to "official" builds, the static package can approach 3 GB due to bigger object files containing intermediate language code. Downloading such big files is tedious on slower connections and also Amazon intermittently closes long connections, making developer's life unbearable.

`tar.bz2` seems to be the best compression method readily available in Python 2.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,14 +24,14 @@ pipeline {
                 }
                 withCredentials([string(credentialsId: 'libccbucket', variable: 'LIBCC_BUCKET')]) {
                   withAWS(credentials:'libccs3',region:'us-east-1') {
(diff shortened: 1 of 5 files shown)


ckerr:
Ah, I see now xz wasn't added until later in Python 3. :-1:

alexeykuzmin:
Agree with Charles. In a perfect world the change from zip to tar would be a one line change.
I vote for avoiding mentioning tar in method/variables names.

alespergl:
I was going to use `create_archive` instead of `create_tar` but thought it was too similar to `create_debug_archive` which is a variable. But since you noticed... :)

PR #451, opened 2018-02-12
Enables mips64el build for Chrome 62

zcbenz:
This PR updates the patches of mips64el.


ckerr:
Not sure why this isn't being picked up? `debian-sysroot-image-creator` v0.8.1 uses `http://ftp.jp.debian.org/debian/pool/main/a/at-spi2-core/libatspi2.0-dev_2.22.0-6+deb9u1_mips64el.deb`, whose contents include `./usr/lib/mips64el-linux-gnuabi64/pkgconfig/atspi-2.pc`

zcbenz:
@ckerr Currently for mips64el build we are using the sysroot image provided by Loongson developers, I'm not sure if the Loongson toolchains are compatible with the binaries from Debian Stretch, the Debian's website [seems to say so](https://wiki.debian.org/MIPSPort). I'll confirm with them.

zcbenz:
I got confirmed that we can just use the sysroot images from Chromium, I'll update this branch to get one less patch.

zcbenz:
So everything should be working except for the GCC 4.9 being too old to compile Chromium, which now uses C++14 standard. It would be a long time to get updated toolchain for mips64el, I'm going to disable the mips64el builds and merge the patches first.

PR #452, opened 2018-02-13
add patch to fix catalog vuln

codebytere:
Add patch based on changes from 
https://chromium-review.googlesource.com/c/chromium/src/+/804397 
to fix 
https://bugs.chromium.org/p/project-zero/issues/detail?id=1450

/cc @alexeykuzmin @deepak1556

### DIFF
--- a/patches/076_patch_catalog_vuln.patch
+++ b/patches/076_patch_catalog_vuln.patch
@@ -47,20 +47,3 @@ index f25a789..2d984f8 100644
    spec.provides["catalog:catalog"].insert("catalog::mojom::Catalog");
    spec.provides["control"].insert("catalog::mojom::CatalogControl");
    InterfaceProviderSpecMap specs;
-diff --git a/mash/quick_launch/manifest.json b/mash/quick_launch/manifest.json
-index 8dfbf4a..ac63cbe 100644
---- a/mash/quick_launch/manifest.json
-+++ b/mash/quick_launch/manifest.json
-@@ -7,7 +7,10 @@
-      "provides": { "mash:launchable": [ "mash::mojom::Launchable" ] },
-      "requires": {
-        "*": [ "app", "mash:launchable" ],
--        "catalog": [ "catalog:catalog" ]
-+        "catalog": [
-+          "catalog:catalog",
-+          "directory"
-+        ]
-      }
-    }
-  }
-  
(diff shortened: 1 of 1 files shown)


alexeykuzmin:
All the other changed `manifest.json` files do not exist in the tree, right?

codebytere:
nope, i removed the ones that failed with `file not found`

alexeykuzmin:
[mash/quick_launch/manifest.json](https://chromium.googlesource.com/chromium/src/+/61.0.3163.100/mash/quick_launch/manifest.json) and [services/ui/manifest.json](https://chromium.googlesource.com/chromium/src/+/61.0.3163.100/services/ui/manifest.json) are there.
I'm not sure if we use them though.

codebytere:
for some reason it's failing to apply on `mash/quick_launch/manifest.json `, i added `services/ui/manifest.json` just now

alexeykuzmin:
Btw, can you please add the patch description to the patch file itself to make @alespergl happy? )

patches/076_patch_catalog_vuln.patch
```diff
@@ -0,0 +1,66 @@
...
+   InterfaceProviderSpec spec;
+-  spec.provides["app"].insert("filesystem::mojom::Directory");
++  spec.provides["directory"].insert("filesystem::mojom::Directory");
+   spec.provides["catalog:catalog"].insert("catalog::mojom::Catalog");
+   spec.provides["control"].insert("catalog::mojom::CatalogControl");
+   InterfaceProviderSpecMap specs;
+diff --git a/mash/quick_launch/manifest.json b/mash/quick_launch/manifest.json
+index 8dfbf4a..ac63cbe 100644
+--- a/mash/quick_launch/manifest.json
++++ b/mash/quick_launch/manifest.json
```
deepak1556: We don't use mash services.

deepak1556:
> mash/quick_launch/manifest.json and services/ui/manifest.json are there. I'm not sure if we use them though.

@alexeykuzmin content browser uses //services/ui https://cs.chromium.org/chromium/src/content/public/app/mojo/content_browser_manifest.json?dr&l=99 and mash service are used only for chromeOS at the moment.

alexeykuzmin:
rel electron/electron#11916

PR #464, opened 2018-03-02
Enable most V8 DCHECKs

alespergl:
It seems that only a few V8 DCHECKs still fail, so I'm enabling the rest.


alexeykuzmin:
@alespergl These changes have already been merged to the libcc's master, right?

alexeykuzmin:
Just rebased it onto the tip of the `electron-2-0-x` branch to mitigate Electron build failure on arm.

PR #468, opened 2018-03-05
Light refactoring of patches related code

alexeykuzmin:
The intention is to move all patches related meta info (owner, description, etc.) in a dedicated config.
It will also help us to avoid patches' name and position collisions.

This PR is just a preparation for the bigger change and only includes the following:
- create Patch, PatchesList, and PatchesConfig classes
- use Patch and PatchesList in patch.py script
- slightly improve apply-patches script

/cc @zcbenz, @alespergl

### DIFF
--- a/script/apply-patches
+++ b/script/apply-patches
@@ -16,18 +16,20 @@ SRC_DIR = os.path.join(SOURCE_ROOT, SRC)
 
 
 def main():
-  error = apply_patches_for_dir(PATCHES_DIR)
-  if not error:
-    if os.environ.has_key('MAS_BUILD'):
-      error = apply_patches_for_dir(PATCHES_MAS_DIR)
-    elif sys.argv[1] == 'mips64el':
-      error = apply_patches_for_dir(PATCHES_MIPS64EL_DIR)
-    if not error:
-      return
+  folders = [PATCHES_DIR]
+  if os.environ.has_key('MAS_BUILD'):
+    folders.append(PATCHES_MAS_DIR)
+  if sys.argv[1] == 'mips64el':
+    folders.append(PATCHES_MIPS64EL_DIR)
 
-  sys.stderr.write(error + '\n')
-  sys.stderr.flush()
-  return 1
+  for folder in folders:
+    error = apply_patches_for_dir(folder)
+    if error:
+      sys.stderr.write(error + '\n')
+      sys.stderr.flush()
+      return 1
+
+  return 0
 
 
 def apply_patches_for_dir(directory):
@@ -37,7 +39,8 @@ def apply_patches_for_dir(directory):
 
     (success, failed_patches) = apply_patches_from_directory(repo, root)
     if not success:
-      return '{0} failed to apply'.format(os.path.basename(failed_patches[0]))
+      patch_path = failed_patches[0].get_file_path()
+      return '{0} failed to apply'.format(os.path.basename(patch_path))
 
 
 if __name__ == '__main__':
--- a/script/lib/patches.py
+++ b/script/lib/patches.py
@@ -1,0 +1,80 @@
+# import yaml  # TODO: Use for PatchesConfig.
+
+from git import apply as git_apply
+
+
+class Patch:
+  def __init__(self, file_path, repo_path):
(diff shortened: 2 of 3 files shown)


script/apply-patches
```diff
@@ -16,18 +16,20 @@ SRC_DIR = os.path.join(SOURCE_ROOT, SRC)
...
-    if os.environ.has_key('MAS_BUILD'):
-      error = apply_patches_for_dir(PATCHES_MAS_DIR)
-    elif sys.argv[1] == 'mips64el':
-      error = apply_patches_for_dir(PATCHES_MIPS64EL_DIR)
-    if not error:
-      return
+  folders = [PATCHES_DIR]
+  if os.environ.has_key('MAS_BUILD'):
+    folders.append(PATCHES_MAS_DIR)
+  if sys.argv[1] == 'mips64el':
```
ckerr: _(trivial?)_ this is an elif in the previous version. Is it intentional that we could potentially append both MAS and MPS64EL patch directories?

script/apply-patches
```diff
@@ -37,7 +39,8 @@ def apply_patches_for_dir(directory):
 
     (success, failed_patches) = apply_patches_from_directory(repo, root)
     if not success:
-      return '{0} failed to apply'.format(os.path.basename(failed_patches[0]))
+      patch_path = failed_patches[0].get_file_path()
+      return '{0} failed to apply'.format(os.path.basename(patch_path))
```
ckerr: I don't understand the intent of this change.

What's the purpose of `get_file_path()` here when the path is stripped back out on the next line?

script/lib/patches.py
```diff
@@ -0,0 +1,80 @@
...
+
+    return (all_patches_applied, failed_patches)
+
+  def reverse(self, stop_on_error=True):
+    return self.apply(reverse=True, stop_on_error=stop_on_error)
+
+
+class PatchesConfig:
+  def __init__(self, path):
+    raise
```
ckerr: I'm not familiar with this idiom; is this a way of marking the class as a WIP?

zcbenz:
> The code that's here looks fine, but it looks like a WIP at the moment. Are you sure you want to merge this now?

I understand this PR as part of a series of big refactors, separated to make review easier.

script/apply-patches
alexeykuzmin: > we could potentially append both MAS and MPS64EL patch directories

It won't make any sense, but I think it not that important to be explicitly expressed in the code.

I try to write code in a way the make future diffs clearer and code reviews easier.
Like here, if there were an `elif` and I then someone would like to remove the first condition, they would also have to replace `elif` with `if`. Independent conditional statements make diffs clearer:
```diff
-if sys.platform == 'win32':
-  NINJA = '{0}.exe'.format(NINJA)
-elif IS_MIPS64EL_HOST:
+if IS_MIPS64EL_HOST:
   NINJA = os.path.join(VENDOR_DIR, 'depot_tools', 'ninja-build')
```

```diff
-if sys.platform == 'win32':
-  NINJA = '{0}.exe'.format(NINJA)
 if IS_MIPS64EL_HOST:
   NINJA = os.path.join(VENDOR_DIR, 'depot_tools', 'ninja-build')
```

Unnecessary changes also screw up git annotations.

script/apply-patches
alexeykuzmin: Before this PR `failed_patches[0]` was a path to a patch file (a string), but now it's a `Patch` instance. To get its path I had to call `.get_file_path()`.

script/lib/patches.py
alexeykuzmin: Maybe ) I almost never use Python.
The whole `PatchesConfig` class wan't expected to be reviewed here, I should have make it clear. Or even exclude it from the PR altogether.

alexeykuzmin:
> I understand this PR as part of a series of big refactors, separated to make review easier.

Exactly.
There only two options how huge PRs are reviewed: they are either approved blindly, without any actual review ("it seems like this guy knows that he's doing... yeah, ok") or they stay unreviewed for days and weeks.

PR #470, opened 2018-03-06
Store patches meta info in patches configs

alexeykuzmin:
Use a patches config to defines order of patches,
specify a patch owner, and provide a description for a patch.

It should help us to avoid order collisions, make merge conflicts explicit,
and help find owners since git blame might not be reliable
because of frequent changes during Chromium upgrades.


alexeykuzmin:
@alespergl 

> Separating the patch and its description - what's the benefit?

1. To avoid conflicts when a description and a patch contents are changed in different branches at the same time.
2. Make it possible to rewrite the whole patch file with a new diff like this `git diff some/file.c some/another_file.c > ../patches/some.patch`. I do it quite often.

> Introducing the pyyaml dependency - can't we use some already supported data format, e.g. Python's own dictionary syntax?

YAML is more human friendly.
Are you concerned that Microsoft won't be able to use pyyaml internally?

> Why is filename ordering not sufficient?

1. When two patches with the same position are added in different branches it does not cause conflicts right now, but it should.
2. Removing a patch creates a gap in position numbers, and it's annoying.

alexeykuzmin:
All builds were green in https://github.com/electron/electron/pull/12367.

ckerr:
File format debates are mostly personal preference and I don't really care what we use. To the small extent that I care, yaml is fine.

The old file numbering scheme worked but I don't mind using another mechanism instead.

For me the big win is the extra metainfo that gives context of why we have a patch, who the "owner" is, when the patch should be dropped, etc etc.. :+1:

alexeykuzmin:
> followup patch that removes the `001-` prefixes from the filenames

I didn't do it in this PR only to reduce noise and make proposed changes easier to review.

alespergl:
I agree with removing the prefixes, obviously.

The file format is not what concerns me, it's the cost of having another dependency (both inside MS and in every-day workflow).

Let's see how this new model works. We can adjust if need be.

alexeykuzmin:
>  followup patch that removes the 001- prefixes from the filenames

Here it is #485

PR #472, opened 2018-03-07
Fix touchpad lagging issue. port from crrev.com/c/867070

alexeykuzmin:
(cherry picked from commit c74ba9786dbb43894d2bb6ae614ab3fe10942068)

/cc @zcbenz

### DIFF
--- a/patches/077-fix-touchpad-lagging-issue.patch
+++ b/patches/077-fix-touchpad-lagging-issue.patch
@@ -1,0 +1,28 @@
+Fix touchpad lagging issue. port from crrev.com/c/867070
+---
+
+diff --git a/gpu/ipc/in_process_command_buffer.cc b/gpu/ipc/in_process_command_buffer.cc
+index f5ac52f17443..414a7f647028 100644
+--- a/gpu/ipc/in_process_command_buffer.cc
++++ b/gpu/ipc/in_process_command_buffer.cc
+@@ -1026,6 +1026,8 @@ void InProcessCommandBuffer::DidCreateAcceleratedSurfaceChildWindow(
+     SurfaceHandle parent_window,
+     SurfaceHandle child_window) {
+   ::SetParent(child_window, parent_window);
++  ::SetWindowPos(child_window, HWND_BOTTOM, 0, 0, 0, 0,
++                 SWP_NOMOVE | SWP_NOSIZE);
+ }
+ #endif
+ 
+diff --git a/ui/gfx/win/rendering_window_manager.cc b/ui/gfx/win/rendering_window_manager.cc
+index a0528ecf9ab7..72ef021bedbe 100644
+--- a/ui/gfx/win/rendering_window_manager.cc
++++ b/ui/gfx/win/rendering_window_manager.cc
+@@ -49,6 +49,7 @@ void RenderingWindowManager::DoSetParentOnChild(HWND parent) {
+   }
+ 
+   ::SetParent(child, parent);
++  ::SetWindowPos(child, HWND_BOTTOM, 0, 0, 0, 0, SWP_NOMOVE | SWP_NOSIZE);
+ }
+ 
+ void RenderingWindowManager::UnregisterParent(HWND parent) {


bpasero:
@alexeykuzmin @zcbenz @ckerr has this fix been verified in Electron? I ran a test build with VS Code including this fix and I am getting mixed feedback: https://github.com/Microsoft/vscode/issues/13612#issuecomment-371899842

It seems like it only fixes this issue as long as users are not minimizing and restoring the Window :-/

bpasero:
@alexeykuzmin yes, more people confirm that this fix does not help when minimizing/restoring the window. A user confirmed however that the issue is fixed in Chrome 66.

Could it be that the fix is a) not complete or b) requires other changes? When I look at https://chromium-review.googlesource.com/c/chromium/src/+/867070/8/ui/gfx/win/rendering_window_manager.cc#44 I see a second change (around line 53) that does not seem to be part of this backport.

bpasero:
+ @chaopeng on https://github.com/electron/libchromiumcontent/pull/472#issuecomment-372008994 and https://github.com/electron/libchromiumcontent/pull/472#issuecomment-372018020

chaopeng:
I have comfirm this fixed on 1.8.3.
To comfirm this issue fixed or new issue appear:
1. open spyxx
2. open electron, do not resize, minimize
3. refresh spyxx (F5), click find window, choose electron
4. See if the order of windows like the screenshot

screen shot 2018-03-11 at 10 25 21

If yes then the fix is applied.

bpasero:
@chaopeng thanks, I can confirm that this fix does not seem to do the right thing on Electron 1.7.x.

**VSCode with Electron 1.7.9 (before minimize and after restore):**
1 7 9_before_and_after_minmize

**VSCode with Electron 1.7.12 + this backport (before minimize)**
1 7 12_before_minimize

**VSCode with Electron 1.7.12 + this backport (after restore)**
1 7 12_after_minimize

So it seems like something in Chrome 59 is required to make this work properly?

chaopeng:
@bpasero It seems like another fixed issue causing the situation you are seeing. I double checked  rendering_window_manager and crbug log, but I don't  know what caused it. It seems Legacy window does not add back to root window correctly. Could you please also test 1.8.3 to check if it correct on the same machine?

bpasero:
@chaopeng I am seeing this issue still with VS Code using Electron 2.0.x, you can play around with that version from here: https://az764295.vo.msecnd.net/insider/42884a0236151873430adb05f8dd7d2fb32f4ed1/VSCode-win32-x64-1.22.0-insider.zip

When I minimize/restore the window, scrolling is laggy again.

chaopeng:
I can also reproduce your situation on my machine. I found the missing window goes to the hidden window. Chromium has logic when user hide a window, Chromium reparent it to hidden window. Then reparent it back when restore, But this case seems it does not reparent back.

But I can not reproduce it on Chromium and Electron 1.8.3. Is it possible it caused by another patch your using?

Sorry I don't have a machine can build Electron at this time. If you still having trouble, I will setup a machine to try it.

bpasero:
@chaopeng thanks for trying it out. Did you try to reproduce with Electron 2.0.x? 

Any comment on this one:

> Could it be that the fix is a) not complete or b) requires other changes? When I look at https://chromium-review.googlesource.com/c/chromium/src/+/867070/8/ui/gfx/win/rendering_window_manager.cc#44 I see a second change (around line 53) that does not seem to be part of this backport.

?

bpasero:
I have serious doubts that this backport is clever given it is only partially applied. There is a portion of the patch in [`RenderingWindowManager::RegisterChild`](https://chromium-review.googlesource.com/c/chromium/src/+/867070/8/ui/gfx/win/rendering_window_manager.cc#b25) which is not applied in Electron because that method seems to be missing some of the changes that came in after Chrome 61.

chaopeng:
@bpasero I just checked electron v2.0.0-beta.5 which included my patch, it does not have the issue you are seeing.

bpasero:
@chaopeng you can use our [VSCode build](https://az764295.vo.msecnd.net/exploration/b1513227ace6929e278c5d531250bc1d0cf89e2f/VSCode-win32-x64-1.22.0-exploration.zip) with Electron 2.0.0-beta.5 to see the issue is still there after you minimize and restore from the taskbar.

chaopeng:
@bpasero Do you use the offical electron to build vscode? https://github.com/electron/electron/releases/tag/v2.0.0-beta.5
Because I don't see this issue on 2.0.0-beta.5. But I do see the issue on your build.

I did suggest:
1. Test with electron defalut app. See if the issue from electron or vscode.
2. If issue from electron, compare the electron with offical electron.
3. Some Chromium code may be good point to check:

views::NativeViewHost::Detach
views::NativeViewHostAura::NativeViewDetaching
views::NativeViewHostAura::RemoveClippingWindow
aura::Window::Hide

These methods call in minimize and the restore methods are arround. I think something prevent the window attaching in restore.

bpasero:
@chaopeng yes we are using the official Electron release for that build. The only difference maybe that we rename Electron.exe to Code.exe and we set a different icon. I hope that would not have an impact because pretty much every product on top of Electron does that?

PR #473, opened 2018-03-08
Do not use system freetype for arm64

zcbenz:
Might be responsible for https://github.com/electron/electron/issues/12057, even if it does not fix the crash, it is still a good thing to align the config of arm64 with other architectures.


zcbenz:
It is working!

PR #474, opened 2018-03-08
 Do not use system freetype for arm64 (2-0-x)

zcbenz:
Backports #473 to 2-0-x, refs https://github.com/electron/electron/issues/12057.


jkleinsc:
I just tested this change against 2-0-x on our ARM CI and everything is green:
https://arm64-ci.electronjs.org/blue/organizations/jenkins/Electron%20ARM64%2Felectron/detail/arm64-ci-for-2-0-x/3/pipeline

PR #483, opened 2018-03-21
Work to fix libcc unit tests

nitsakh:
This PR makes changes to address libcc unit test issues. (#360 and #443)


nitsakh:
I've just fixed the content_browsertests build and the existing test right now. But we should probably add tests to test our patch code.

alexeykuzmin:
Are these changes related to any of our patches?
If yes, they should be done in the same patch.

alexeykuzmin:
Can you please rebase your branch onto the latest master?
#470 was merged several hours ago and it's added a new config for patches, so when you add a new patch it also has to be added to that config to be actually applied.
I'll update the documentation soon.

deepak1556:
> Are these changes related to any of our patches?
If yes, they should be added to that patch.

Second that, will be easy to reason the changes.

alexeykuzmin:
Also, can you please add your Slack email to your Github account?
GitHub doesn't know that "nitsakh@slack-corp.com" is you at the moment )

nitsakh:
I should've committed with the other email address. Will amend the commit to change the email address.
Changes are related to another patch, but I was thinking that keeping test changes separate could be a good idea.

alexeykuzmin:
> Changes are related to another patch, but I was thinking that keeping test changes separate could be a good idea.

Reasons?

nitsakh:
Yeah, that was the initial thought, but having them in the same patch is the right thing to. I've updated the patch.

nitsakh:
Following tests are also failing after running the run_tests script.

* components_unittests:
	* AutofillManagerTest.UploadCreditCardAndSaveCopy
* accessibility_unittests:
	* AXTreeTest.EmptyNodeOffscreenWhenAllChildrenOffscreen
	* AXTreeTest.GetBoundsUpdatesOffscreen
* views_unittests:
	* WidgetTest.ChildStackedRelativeToParent
* extensions_browsertests:
	* SocketsUdpApiTest.SocketsUdpExtension
* pdfium_unittests:
	* cpdf_parser.RebuildCrossRefCorrectly
	* cpdf_parser.RebuildCrossRefFailed
* shell_dialogs_unittests
* rlz_unittests:
	* RlzLibTest.SendFinancialPing
* compositor_unittests:
	All tests in LayerWithDelegateTest, LayerWithNullDelegateTest and LayerWithRealCompositorTest. 	(@alexeykuzmin has verified that these are failing in chromium as well)
command_buffer_gles2_tests:
	* EGLThreadTest.Basic (This one is similar to the WeakPtr test failure; related to disable DCHECK patch)
* ffmpeg_regression_tests:
	104 test failures. I looked at a few in the list and looks like they’re all because they could not find the data file.

Maybe some of these are not needed, but these were all the failures I got after running the `run_tests` script.  We should probably remove some targets if we do not need them.

patches/common/chromium/dcheck.patch
```diff
@@ -224,3 +224,26 @@ index 674b0e9a909c..a1bff6e40f56 100644
...
++++ b/base/memory/weak_ptr_unittest.cc
+@@ -565,7 +565,9 @@ TEST(WeakPtrTest, NonOwnerThreadCanDeleteWeakPtr) {
+   background.Start();
+   background.DeleteArrow(arrow);
+ }
+-
++/*Disabling the WeakPtrDeathTest tests until WeakReference::Flag::IsValid() 
++DCHECK is fixed*/
++/*
+ TEST(WeakPtrDeathTest, WeakPtrCopyDoesNotChangeThreadBinding) {
```
deepak1556: Can you rename the test desc with `DISABLED_` similar to other disabled tests instead of commenting it out. Thanks!

deepak1556:
Its weird that `extensions_browsertests` got built, there shouldn't be a target in our test build config that has it as a dep, it should be disabled if its getting built. I am good with having rest of the components, as we use parts of them in Electron. We can disable the tests that can't be fix at the moment.

alexeykuzmin:
Conflicts.

patches/common/chromium/.patches.yaml
```diff
@@ -331,3 +331,7 @@ patches:
   owners: zcbenz
   file: fix-arm64-linking-error.patch
   description: Do not use system freetype for arm64
+-
+  owners: nitsakh
+  file: disable_unittests.patch
+  description: Disabling unit tests
```
alexeykuzmin: New line please _

patches/common/chromium/disable_unittests.patch
```diff
@@ -0,0 +1,14 @@
+Disable GPU config test as it fails on High Sierra
```
alexeykuzmin: It would probably be better to list all disabled tests and the reasons why they're disabled in the `.patches.yaml`.

nitsakh: Sure. Makes sense. Will update.
I've also added comments above some disabled test in the patch file.

alexeykuzmin:
Let's merge it. CI builds are green, so at least you didn't break anything (yet) =)

nitsakh:
How did I manage to break all of CI? 🤔

alexeykuzmin:
@nitsakh 

> disable-warning-win.patch failed to apply

alexeykuzmin:
@nitsakh It seems like Windows build are OK. Should we merge the PR?

nitsakh:
Yes, let's merge the PR. We can then get this in CI. If there are some things, we can work on those after the merge.

PR #486, opened 2018-03-22
Fix bootstrap install error

nitsakh:
I was seeing the following error on macOS when running the bootstrap script.
`error: can't combine user with prefix, exec_prefix/home, or install_(plat)base`
This fixes it.


alexeykuzmin:
If CI bots are happy then we should probably be too )

PR #487, opened 2018-03-22
add patch to disable base_computed_style_ DCHECK

codebytere:
This `DCHECK` is for some reason causing failures in `electron-api-demos` for reasons that i've investigated but can't fully determine; i'm disabling this check for now as a mitigation and will look to more fully investigate and re-add the check in later.

/cc @ckerr


ckerr:
Fixes https://github.com/electron/electron/issues/11949.

deepak1556:
The patch should be added to https://github.com/electron/libchromiumcontent/blob/master/patches/common/chromium/dcheck.patch . Also it can be just

```
#if 0
  if (base_computed_style_ && computed_style)
    DCHECK(*base_computed_style_ == *computed_style);
#endif
```

for minimal changes. Thanks!

PR #489, opened 2018-03-23
[2-0-x]: Re-enable accelerated 2D Canvas on NVidia High Sierra

deepak1556:
As per https://bugs.chromium.org/p/chromium/issues/detail?id=773705#c141, I am aligning the patch with upstream. Might improve the state of https://github.com/electron/electron/issues/12042

```
Re-enabling accelerated 2D Canvas, as without acceleration performance
was bad enough that Canvas2D became unusable in common cases. If
deciding between occasional visual corruption and unusable performance,
visual corruption seems to be the lesser issue.

Note that this keeps GPU Raster blacklisted. This is unfortunate, but
was only enabled 1yr ago, so should have less sites with a hard
dependency.
```

CL: https://chromium-review.googlesource.com/c/chromium/src/+/759236

/cc @bpasero @Tyriar

### DIFF
--- a/patches/043-fix_rendering_glitches_on_high_sierra.patch
+++ b/patches/043-fix_rendering_glitches_on_high_sierra.patch
@@ -1,8 +1,8 @@
 diff --git a/gpu/config/software_rendering_list.json b/gpu/config/software_rendering_list.json
-index ed5c6d6c0e3c..ffec922b199d 100644
+index ed5c6d6c0e3c..da2053366a07 100644
 --- a/gpu/config/software_rendering_list.json
 +++ b/gpu/config/software_rendering_list.json
-@@ -1532,6 +1532,24 @@
+@@ -1532,6 +1532,23 @@
        "features": [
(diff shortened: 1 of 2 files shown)


MarshallOfSound:
Windows died for what appears to be an unrelated reason 🤔

ckerr:
Keeps dying on 

```text
crashpad/util/win/nt_internals.cc /Foobj/third_party/crashpad/crashpad/util/util/nt_internals.obj /Fd"obj/third_party/crashpad/crashpad/util/util_cc.pdb"
../../third_party/crashpad/crashpad/util/win/nt_internals.cc(22): error C2371: 'CLIENT_ID': redefinition; different basic types
c:\program files (x86)\windows kits\10\include\10.0.16299.0\um\winternl.h(83): note: see declaration of 'CLIENT_ID'
```

Unrelated or not, this seems like something that needs to get solved if we want to have this fix in 2.0.0-beta.6

PR #491, opened 2018-03-23
move DCHECK disable to DCHECK patchfile

codebytere:
This PR moves the DCHECK disabled in https://github.com/electron/libchromiumcontent/pull/487 to the amalgamate file `dcheck.patch`

/cc @deepak1556


patches/common/chromium/dcheck.patch
```diff
@@ -3,12 +3,12 @@ index 582fb89868cd..24d3e82232b6 100644
 --- a/base/logging.h
 +++ b/base/logging.h
 @@ -850,7 +850,7 @@ const LogSeverity LOG_DCHECK = LOG_INFO;
- 
+
```
alexeykuzmin: All those whitespace shouldn't have been removed =/
@codebytere

codebytere: yeah i know, i have literally no clue how that happened but i'll fix it

PR #493, opened 2018-03-26
Add patch to fix duplicate client id define

codebytere:
Fixes blocker in  https://github.com/electron/libchromiumcontent/pull/489.

This patch removed the define in Chromium to mitigate conflicts with Windows SDK's define.

/cc @ckerr @zcbenz @deepak1556


deepak1556:
Have backported to `2-0-x` here https://github.com/electron/libchromiumcontent/pull/489/commits/96084c6a3758281d8a2e658d4860552d955aaff9

deepak1556:
The issue is fixed with building on an older sdk following upstream changes https://bugs.chromium.org/p/chromium/issues/detail?id=773476 https://github.com/electron/libchromiumcontent/pull/489/commits/3ad32ab74bd9889c94825df3a8fe3908e109e06b. The same fix is already available in master, so the crashpad patch need not be backported. Thanks!

alexeykuzmin:
@codebytere Based on the comment above, should we close this PR? )

PR #499, opened 2018-03-27
Chromium 66

alexeykuzmin:
Project board: https://github.com/electron/electron/projects/10
Electron PR: https://github.com/electron/electron/pull/12477

PR #502, opened 2018-03-31
Remove guest_view.patch

codebytere:
There is a new video capture API that can do everything needed, so this patch is no longer required. 
Discussed with @brenca, who will push updates to the `electron/electron` side to compensate for new API. 

For https://github.com/electron/libchromiumcontent/pull/499.


alexeykuzmin:
> Discussed with @brenca, who will push updates to the electron/electron side to compensate for new API.

@brenca Is there a task or an open PR for this?
I don't want to accidentally forget about it.

brenca:
@alexeykuzmin I see you added a task in the project already, thanks! I'll take a stab at this as soon as electron can be built with C66.

alexeykuzmin:
@brenca Cool, thank you!

PR #503, opened 2018-04-02
Add a patch to fix window repainting issues when DWM is disabled on Windows

brenca:
Fixes electron/electron#1821, sister PR coming up on that repo

When DWM is disabled, chromium uses a custom window title implementation (this can be observed by running Chrome on Windows7 with the aero theme disabled). The code is written to assume that that's what's happening, but electron is forcing a native titlebar, and that causes problems, since chromium does [some hacky stuff](https://cs.chromium.org/chromium/src/ui/views/win/hwnd_message_handler.cc?l=253) to ensure that the custom titlebar always works perfectly. That hacky stuff blocks the native titlebar (and the whole window in fact) from rendering correctly. This patch adds a way to disable the lock.

### DIFF
--- a/patches/common/chromium/.patches.yaml
+++ b/patches/common/chromium/.patches.yaml
@@ -274,10 +274,6 @@ patches:
   file: disable-recursive-surface-sync.patch
   description: null
 -
-  owners: brenca
-  file: disable-redraw-lock.patch
-  description: null
--
   owners: alexeykuzmin
   file: backport_a6977980088b.patch
   description: null
@@ -335,3 +331,17 @@ patches:
   owners: zcbenz
   file: fix-arm64-linking-error.patch
   description: Do not use system freetype for arm64
+-
+  owners: brenca
+  file: disable-redraw-lock.patch
+  description:  |
+    Chromium uses a custom window titlebar implementation on Windows when DWM
+    is disabled (Windows 7 and earlier, non Aero theme). The native titlebar
+    sometimes painted over this custom titlebar, so a workaround was put in
+    place to lock redraws in reaction to certain events if DWM is disabled,
+    since the code assumes that in that case, the custom titlebar is painted.
+    Electron forces the use of the native titlebar, which the workaround doesn't
+    take into account, and still locks redraws, causing weird repainting issues
+    in electron (and other applications). This patch provides a way to disable
+    the redraw locking mechanism, which fixes these issues. The electron issue
+    can be found at https://github.com/electron/electron/issues/1821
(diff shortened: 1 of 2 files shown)


brenca:
@alexeykuzmin Should I add a description to `patches.yaml` too?

patches/common/chromium/.patches.yaml
```diff
@@ -273,6 +273,10 @@ patches:
   owners: deepak1556
   file: disable-recursive-surface-sync.patch
   description: null
+-
+  owners: brenca
+  file: disable-redraw-lock.patch
```
alexeykuzmin: Any reasons to add it in the middle of the patches list, and not at the end?

patches/common/chromium/.patches.yaml
```diff
@@ -179,7 +179,7 @@ patches:
     Linux, which may cause issues in a context like Electron, where the main
     and renderer threads are supposed to keep inherited permissions over the
     system.
-    
+
```
alexeykuzmin: Thanks! )

patches/common/chromium/.patches.yaml
```diff
@@ -273,6 +273,10 @@ patches:
   owners: deepak1556
   file: disable-recursive-surface-sync.patch
   description: null
+-
+  owners: brenca
+  file: disable-redraw-lock.patch
+  description: null
```
alexeykuzmin: It would be better to have an explanation why it is needed.
For instance a link to an issue it fixes, or a feature it implements.

alexeykuzmin:
>  Should I add a description to .patches.yaml too?

@brenca 
It would be great. Thanks!

patches/common/chromium/.patches.yaml
brenca: I just tried to maintain alphabetical order :)

patches/common/chromium/.patches.yaml
brenca: That was automatic from my Atom, I hope it doesn't break anything :D

patches/common/chromium/.patches.yaml
brenca: On it!

brenca:
@alexeykuzmin I'm ready for another review :)
BTW this `.patches.yaml` file is a pretty nice addition, it's much easier to see what patch does what this way. (I haven't seen it before, a long time has gone by since I added any patches :) )

There are some other patches that I authored, I can add descriptions in a separate PR for those if you want.

alexeykuzmin:
> There are some other patches that I authored, I can add descriptions in a separate PR for those if you want.

That would be great! But please do it in a separate PR it you don't mind )

alexeykuzmin:
I just restarted the Windows builds, after they finish we should be good to merge.

brenca:
> > There are some other patches that I authored, I can add descriptions in a separate PR for those if you want.
> 
> That would be great! But please do it in a separate PR if you don't mind )

Sure, that was my intention, I don't want to pollute this PR with more changes :)

PR #507, opened 2018-04-04
Cleanup V8 patches

alespergl:
- Remove mksnapshot.patch (shouldn't be necessary anymore)
- Export only `v8::Platform::SystemClockTimeMillis`, not the whole class


patches/common/v8/.patches.yaml
```diff
@@ -64,6 +56,11 @@ patches:
   owners: codebytere
   file: cherry_pick_ac0fe8ec.patch
   description: null
+-
+  owners: alexeykuzmin,alespergl
+  file: export_platform.patch
```
alexeykuzmin: I'm pretty sure you moved the patch closer to the end of the list intentionally )
What's the reason to do that?

patches/common/v8/.patches.yaml
```diff
@@ -64,6 +56,11 @@ patches:
   owners: codebytere
   file: cherry_pick_ac0fe8ec.patch
   description: null
+-
+  owners: alexeykuzmin,alespergl
```
alexeykuzmin: I would ask you to add a space after the comma, but I don't want to wait for new builds to finish )
This file needs a clean up anyway, (I hope that) I'll do it later.

patches/common/v8/.patches.yaml
alespergl: Yes, because I generated the patch from a state where the previous patches are already applied. 😉

patches/common/v8/.patches.yaml
alespergl: Ok, next time I'll know. I was looking for an example of multiple owners in the other yaml files, but couldn't find any, so I left it like this.

alexeykuzmin: As for know, value of the "owners" fields is not used anywhere, so I can be in any format.
It should probably be an array, but it would require usage of a heavy `[ alexeykuzmin, alespergl ]` notation. So I think simple comma-separated values are fine now.

PR #511, opened 2018-04-09
Refactor some scripts

alexeykuzmin:
No breaking changes
(with exception of renamed `--target-arch` → `--target_arch` arg of the `apply-patches` script).

* Move some script constants to a shared "lib/config.py"
* Add and use `git.get_head_commit()`
* Extract Ninja related stuff to "lib/ninja.py"
* Add 'run-gn' script and call it from the 'update' script
* Name an arg "target_arch" instead of "target-arch" for consistency

### DIFF
--- a/script/apply-patches
+++ b/script/apply-patches
@@ -52,7 +52,7 @@ def apply_patches_for_dir(directory):
 def parse_args():
   parser = argparse.ArgumentParser(description='Apply all required patches.')
 
-  parser.add_argument('-t', '--target-arch',
+  parser.add_argument('-t', '--target_arch',
                       help='Target architecture')
 
   return parser.parse_args()
--- a/script/lib/ninja.py
+++ b/script/lib/ninja.py
@@ -1,0 +1,26 @@
+import os
+import subprocess
+import sys
+
+from lib.config import DEPOT_TOOLS_DIR
+
+
+def __get_binary_path():
+  path = os.path.join(DEPOT_TOOLS_DIR, 'ninja')
+
+  if sys.platform == 'win32':
+    path = '{0}.exe'.format(path)
+
+  return path
+
+
+def run(directory, target=None, env=None):
+  ninja_binary = __get_binary_path()
+
+  args = [ninja_binary,
+          '-C', directory
+         ]
+  if target is not None:
+    args.append(target)
+
+  subprocess.check_call(args, env=env)
--- a/script/lib/patches.py
+++ b/script/lib/patches.py
@@ -1,8 +1,10 @@
 import os
 import sys
 
-PROJECT_ROOT = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
-sys.path.append(os.path.join(PROJECT_ROOT, 'vendor', 'pyyaml', 'lib'))
+from config import VENDOR_DIR
+
+PYYAML_LIB_DIR = os.path.join(VENDOR_DIR, 'pyyaml', 'lib')
+sys.path.append(PYYAML_LIB_DIR)
 import yaml
 
 from git import apply as git_apply
--- a/script/run-gn
+++ b/script/run-gn
@@ -1,0 +1,85 @@
+#!/usr/bin/env python
+
+import argparse
+import json
+import os
+import sys
+
+import lib.gn as gn
+
+from lib.config import get_output_dir, COMPONENTS, \
+                       SOURCE_ROOT, SRC_DIR, DEPOT_TOOLS_DIR
+
+
+# TODO(alexeykuzmin): There are no reasons to keep args files
+# in the "chromiumcontent" folder.
+CHROMIUMCONTENT_SOURCE_DIR = os.path.join(SOURCE_ROOT, 'chromiumcontent')
+
+
+def parse_args():
+  parser = argparse.ArgumentParser(description='Generate Ninja')
+
+  parser.add_argument('-a', '--args', type=json.loads, default=None,
+                      help='List of GN args in JSON format. E.g. \'{"key": "value"}\'')
+  parser.add_argument('-c', '--components', nargs='+', default=COMPONENTS, choices=COMPONENTS,
+                      help='Component(s) to generate build files for.')
+  parser.add_argument('-t', '--target_arch', default='x64', help='E.g. "x64".')
+
+  return parser.parse_args()
+
+
+def main():
+  args = parse_args()
+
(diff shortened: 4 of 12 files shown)


script/lib/ninja.py
```diff
@@ -0,0 +1,32 @@
+"""Git helper functions.
```
alexeykuzmin: Oops

script/update
```diff
@@ -252,38 +248,16 @@ def setup_mips64el_toolchain(target_arch):
...
-    "linux2": "linux",
-    "win32": "win",
-  }.get(sys.platform)
 
-  if target_arch == 'ia32':
-    target_cpu = 'x86'
-  else:
-    target_cpu = target_arch
+def run_gn(target_arch):
+  gn_args = {}
```
alexeykuzmin: @MarshallOfSound 
After this change adding `cc_wrapper="sccache"` should be as simple as:

```python
if use_sccache:
  gn_args['cc_wrapper'] = ''sccache"
```

script/apply-patches
```diff
@@ -52,7 +52,7 @@ def apply_patches_for_dir(directory):
 def parse_args():
   parser = argparse.ArgumentParser(description='Apply all required patches.')
 
-  parser.add_argument('-t', '--target-arch',
+  parser.add_argument('-t', '--target_arch',
```
nornagon: this looks like a breaking change, but probably nbd. if you wanted to be extra-cautious, you could:

```python
parser.add_argument('--target-arch', help=argparse.SUPPRESS)
```

script/run-gn
```diff
@@ -0,0 +1,96 @@
...
+  parser.add_argument('-t', '--target_arch', default='x64', help='E.g. "x64".')
+
+  return parser.parse_args()
+
+
+def main():
+  args = parse_args()
+
+  # Generate build files for all components if none are specified.
+  components = args.component if args.component else COMPONENTS
```
nornagon: could this logic be folded into the `default` param of the `--component` arg definition?

script/run-gn
```diff
@@ -0,0 +1,96 @@
...
+  # Add new and override default args from a command line argument.
+  if args.args:
+     gn_args.update(args.args)
+
+  for component in components:
+    generate_build_files(component, target_arch, **gn_args)
+
+
+def generate_build_files(component, target_arch, **gnargs):
+  build_dir = create_build_dir(component, target_arch)
```
nornagon: i don't think GN requires the build dir to exist before running `gn gen`, so you might be able to skip this bit

script/update
```diff
@@ -252,38 +248,16 @@ def setup_mips64el_toolchain(target_arch):
...
-          target_cpu=target_cpu)
-
-    # Generate Ninja build files.
-    gn.generate(output_dir, chromium_root_dir=SRC_DIR, depot_tools_dir=DEPOT_TOOLS, env=env)
+  script_path = os.path.join(SOURCE_ROOT, 'script', 'run-gn')
+  script_arguments = [
+      '--args', json.dumps(gn_args),
+      '--target_arch', target_arch
+      ]
+  return subprocess.call([sys.executable, script_path] + script_arguments)
```
nornagon: why a subprocess rather than a function call?

alexeykuzmin: Who knows ¯\_(ツ)_/¯

script/apply-patches
alexeykuzmin: I hope that nobody actually calls `apply-patches` manually.
Since the change doesn't break the CI build, we should be fine.

script/run-gn
alexeykuzmin: Probably. Thanks!

script/lib/patches.py
```diff
@@ -1,8 +1,10 @@
 import os
 import sys
 
-PROJECT_ROOT = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
-sys.path.append(os.path.join(PROJECT_ROOT, 'vendor', 'pyyaml', 'lib'))
+from config import VENDOR_DIR
+
+PYYAML_LIB_DIR = os.path.join(VENDOR_DIR, 'pyyaml', 'lib')
+sys.path.append(PYYAML_LIB_DIR)
```
ckerr: Small preference to not leave `PYYAML_LIB_DIR` laying around if it's not going to be used elsewher. Maybe `sys.path.append(os.path.join(...))`

script/run-gn
```diff
@@ -0,0 +1,96 @@
...
+  if sys.platform in ['win32', 'cygwin']:
+    env['DEPOT_TOOLS_WIN_TOOLCHAIN'] = '0'
+
+  gn.generate(build_dir, chromium_root_dir=SRC_DIR, depot_tools_dir=DEPOT_TOOLS_DIR, env=env)
+
+
+def create_build_dir(component, target_arch):
+  build_dir = get_output_dir(SOURCE_ROOT, target_arch, component)
+  if not os.path.isdir(build_dir):
+    os.makedirs(build_dir)
```
ckerr: `exist_ok=False` ... oh wait, that's py3

script/apply-patches
alexeykuzmin: > idea for a backwards-compatible option

The arg [was added](https://github.com/electron/libchromiumcontent/pull/470/commits/931f976528e634600048a79787b071a6c9cbc7c3) only three weeks ago, nobody will ever notice the change.
And I don't think build scripts have a public contract anyway )

script/run-gn
alexeykuzmin: Done

script/run-gn
alexeykuzmin: `gn gen` doesn't require the build dir to exist but we write `args.gn` file to the build dir before running `gn gen` for it, so the dir should be there anyway.
Moved dir creation to the `lib/gn.py`.

alexeykuzmin:
@nornagon Any other suggestions to make it even better? Did I forget to change anything?

script/update
nornagon: It would be really easy! Just rename `run-gn` to `run_gn.py`, then `from run_gn import main as run_gn` and change `main` to accept argv as a parameter. Even better would be to factor out the functionality you want to access as a function with sensible arguments and call it, e.g. `def generate_all_build_files(components=COMPONENTS)`.

alexeykuzmin:
@ckerr, @nornagon 
Thanks for the feedback. I'll try to take it into account on the next iteration )

PR #513, opened 2018-04-10
Add flag for enabling sccache

MarshallOfSound:
Use with `--sccache`

### DIFF
--- a/script/update
+++ b/script/update
@@ -74,7 +74,7 @@ def main():
           copy_chromiumcontent_files() or
           update_clang() or
           setup_mips64el_toolchain(target_arch) or
-          run_gn(target_arch, args.sccache))
+          run_gn(target_arch, args.cc_wrapper))
 
 
 def parse_args():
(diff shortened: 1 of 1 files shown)


script/update
```diff
@@ -96,6 +96,7 @@ def parse_args():
                       help='Skips update depot tools on Windows, easier to build locally')
   parser.add_argument('--source_only', action='store_true',
                       help='Only sync the Chromium source code')
+  parser.add_argument('--sccache', default=False, help='Enable sccache support')
```
alexeykuzmin: Shouldn't it have `action='store_true'`?

alexeykuzmin:
> We should expose `cc_wrapper` rather than hardcoding sccache, it might make sense to use some other wrapper/s, e.g. regular ccache.

I think that having `--sccache` as a shortcut for `--cc_wrapper=sccache` is a good thing.

alexeykuzmin:
> I think that having `--sccache` as a shortcut for `--cc_wrapper=sccache` is a good thing.

I mean something like that:
```python
cc_wrapper_group = parser.add_mutually_exclusive_group()
cc_wrapper_group.add_argument('--cc_wrapper', help="...")
cc_wrapper_group.add_argument('--ccache', action='store_const', const='ccache',
                              help="...")
cc_wrapper_group.add_argument('--sccache', action='store_const', const='sccache',
                              help="...")
```

alexeykuzmin:
> We should expose cc_wrapper rather than hardcoding sccache, it might make sense to use some other wrapper/s, e.g. regular ccache.

@nornagon done

PR #514, opened 2018-04-12
Add flag to cache libcc downloads

MarshallOfSound:
Will need a follow up PR to electron to allow usage of this, basically you can optionally cache libcc downloads to a folder on your machine.  Means that when you switch branch you won't have to redownload stuff.

### DIFF
--- a/script/download
+++ b/script/download
@@ -9,11 +9,15 @@ import sys
 import tempfile
 import urllib2
 import tarfile
+import shutil
 
 import lib.filesystem as filesystem
 
 
 SOURCE_ROOT = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
+CACHE_ROOT = os.path.join(tempfile.gettempdir(), 'libcc_cache')
+if 'ELECTRON_LIBCC_CACHE_DIR' in os.environ:
+  CACHE_ROOT = os.environ['ELECTRON_LIBCC_CACHE_DIR']
 
 PLATFORM_KEY = {
   'cygwin': 'win',
@@ -33,6 +37,9 @@ class ProgramError(Exception):
 def main():
   try:
     args = parse_args()
+    if args.cache:
+      print('Using cache root: ' + CACHE_ROOT + '\n')
+      mkdir_p(CACHE_ROOT)
     if (args.libcc_source_path != None and
         args.libcc_shared_library_path != None and
         args.libcc_static_library_path != None):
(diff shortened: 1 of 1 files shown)


alexeykuzmin:
Would it make sense to give unique local names to already downloaded libcc archives to avoid redownload?

MarshallOfSound:
> Would it make sense to give unique local names to already downloaded libcc archives to avoid redownload?

What do you mean, all libcc downloads are stored at `{commit}-{filename}` so they won't redownload 🤔

alexeykuzmin:
> What do you mean, all libcc downloads are stored at `{commit}-{filename}` so they won't redownload 🤔

@MarshallOfSound 
Sorry, I didn't read your changes as carefully as I should have.
That's exactly what you're doing.

script/download
```diff
@@ -162,6 +189,13 @@ def generate_filenames_gypi(target_file, libcc_source_path,
                          libcc_source_path, libcc_shared_library_path,
                          libcc_static_library_path])
 
+def mkdir_p(path):
```
alexeykuzmin: Already exists in `lib/filesystem.py`.

script/download
```diff
@@ -9,11 +9,15 @@ import sys
 import tempfile
 import urllib2
 import tarfile
+import shutil
```
alexeykuzmin: Imports should be listed in the alphabetical order.

script/download
```diff
@@ -80,6 +87,7 @@ def parse_args():
                       help='Overwrite destination if it already exists.')
   parser.add_argument('-c', '--commit', nargs='?', default='HEAD',
                       help='The commit of libchromiumcontent to download.')
+  parser.add_argument('--cache', action='store_true')
```
alexeykuzmin: `--use_cache` maybe? `cache` is slightly weird name for a boolean argument.

script/download
```diff
@@ -124,30 +132,49 @@ def download_if_needed(destination, base_url, commit, filename, force):
...
 
-  download(destination, base_url, commit, filename)
+  download(destination, base_url, commit, filename, use_cache)
 
   with open(version_file, 'w') as f:
     f.write('{0}\n'.format(commit))
 
 
-def download(destination, base_url, commit, filename):
+def download(destination, base_url, commit, filename, use_cache):
```
alexeykuzmin: I would probably move the whole logic here. Like
```python
archive = None
cache_key = ...
url = ...

if use_cache:
  archive = get_from_cache(cache_key)

if archive is None:
  archive = download(url)
  if archive is not None and use_cache:
    save_to_cache(key=cache_key, file=archive)

if archive is not None:
  extract(from=archive, to=destination)
else:
  print 'Better luck next time, mate'
```

PR #516, opened 2018-04-13
Bundle sccache binaries

alexeykuzmin:


alexeykuzmin:
> FIXME: Try to run bundled sccache

`sccache --version` works fine on Windows and Linux, Mac TBD.

MarshallOfSound:
@alexeykuzmin Where did the prebuilt's come from, they need to come from my hashing branch if we want the hashing to work across machines 🤔

alexeykuzmin:
@MarshallOfSound 
https://github.com/mozilla/sccache/releases/tag/0.2.6

alexeykuzmin:
> if we want the hashing to work across machines

Honestly I don't even understand what you're talking about )

alexeykuzmin:
> Mac fails with `dyld: Library not loaded: /usr/local/opt/openssl@1.1/lib/libssl.1.1.dylib` error

Everything's ok now. `sccache` statically linked with the `openssl` works well.

alexeykuzmin:
@alespergl it's 25MB of binaries combined.
I probably could make them entirely optional: like put them in an external storage, add a script to fetch them when they're needed and os on, but it would be too complicated.

I totally understand your concern though. If we continue to add more stuff in the repo, eventually it will become huge and slow, and we definitely don't want it to happen.

PR #521, opened 2018-04-17
Update run_tests script to run only specified test targets

nitsakh:
This will enable us to run only desired test targets.
Allows to include specified targets, or exclude.


alexeykuzmin:
Didn't understand the implementation but like the interface =) Approved.

PR #522, opened 2018-04-17
Use sccache to speed up builds

jkleinsc:
This PR updates our CI to use sccache in order to speed up builds.  
cc #516


jkleinsc:
It appears that our current S3 configuration isn't working due to this issue:
https://github.com/mozilla/sccache/issues/225

Here is an example build where it missed 100% of a warmed cache:
https://circleci.com/gh/electron/libchromiumcontent/6146

MarshallOfSound:
@jkleinsc There is a way around this, pinging you in Slack 👍

.circleci/config.yml
```diff
@@ -14,13 +14,19 @@ jobs:
           command: script/bootstrap
       - run:
           name: Update
-          command: script/update --clean -t $TARGET_ARCH
+          command: script/update --clean -t $TARGET_ARCH --use-bundled-sccache
+      - run:
+          name: Check sccache stats before build
+          command: tools/sccache/0.2.6/linux/sccache -s
```
alexeykuzmin: Maybe we can add a wrapper script to run bundled `sccache`? Like `script/sccache` that would run bundled `sccache` for current platform?
These absolute paths with versions and platforms in them look difficult to maintain.

script/cibuild.ps1
```diff
@@ -33,8 +33,13 @@ function Run-Command([scriptblock]$Command, [switch]$Fatal, [switch]$Quiet) {
...
+$CommandLine = "python .\script\cibuild"
 if ($skipUpload) {
-  Run-Command -Fatal { python .\script\cibuild --skip_upload }
-} else {
-  Run-Command -Fatal { python .\script\cibuild }
+  $CommandLine += " --skip_upload"
 }
+
+if ($useSccache) {
+  $CommandLine += " --use_sccache"
```
alexeykuzmin: Does it work on Windows? )

appveyor.yml
```diff
@@ -1,32 +0,0 @@
-version: 1.0.{build}
```
alexeykuzmin: We don't need it anymore?

script/cibuild.ps1
jkleinsc: This command does, sccache however isn't working on Windows.  I wanted to leave this code in because we will need it once we enable sccache on Windows.

appveyor.yml
jkleinsc: No, we don't need it.  It is just needed when we need a configuration different from the defaults configured on the server.

jkleinsc:
@alexeykuzmin binaries were built from master.  I didn't bother to relocate them since the intention is to not keep them around in the repo but instead provide a way to download them.

alexeykuzmin:
Git shows the linux version of sccache as simply renamed:
`tools/sccache/0.2.6/linux/sccache → tools/sccache/aad2120/linux/sccache`.

Is it true? )

jkleinsc:
@alexeykuzmin 

> Git shows the linux version of sccache as simply renamed:
> tools/sccache/0.2.6/linux/sccache → tools/sccache/aad2120/linux/sccache.
> 
> Is it true? )

No, because if you look at the linux build result here, its using the Azure storage which wasn't available in 0.2.6: https://circleci.com/gh/electron/libchromiumcontent/6413

alexeykuzmin:
@jkleinsc Shell we merge it? )

.circleci/config.yml
```diff
@@ -50,7 +56,7 @@ jobs:
           command: script/bootstrap
       - run:
           name: Update
-          command: script/update --clean -t $TARGET_ARCH
+          command: script/update --clean -t $TARGET_ARCH --use-bundled-sccache
```
alexeykuzmin: Somehow missed it.
I thought we agreed on not using cache for Release builds.

jkleinsc: @alexeykuzmin why would we not use cache for Release builds???

jkleinsc: I mean it doesn't really gain us much if we only use this for shared builds.  We still end up having to wait for the static builds.  Given that we are using a cache that only the CI has write access to, I'm not sure why we wouldn't use it for Release builds.

MarshallOfSound: @jkleinsc For maximum safety, release builds should be generated in a 100% deterministic and isolated way.  Generating them without cache is what everyone was 👍 on from memory

jkleinsc: Alright then, I changed it to just run for shared builds.

PR #526, opened 2018-04-19
Fix PR builds for Mac

jkleinsc:
Visual Studio Team Services builds PRs by merging the commit(s) into the target branch and then building that.  Unfortunately this ends up putting PR builds in the wrong location on S3 as well as it produces issues for WIP PRs such as our Chromium upgrades.

This PR will do the following:
1. If the build is not a PR build, it will build the commit.
2. If the build is a PR build, but for a fork, it will build against the result of merging the commit(s) into the target branch.
3. If the build is a PR build from a branch in this repo, it will build the original commit instead of the merge.

### DIFF
--- a/vsts.yml
+++ b/vsts.yml
@@ -27,7 +27,20 @@ phases:
 
   steps:
   - bash: |
-     echo "##vso[task.setvariable variable=gitcommit]$BUILD_SOURCEVERSION"
+     if [[ -z "${SYSTEM_PULLREQUEST_PULLREQUESTNUMBER}" ]] || "$SYSTEM_PULLREQUEST_ISFORK" == "True"; then
+       echo "##vso[task.setvariable variable=gitcommit]$BUILD_SOURCEVERSION"
+     else
+       tmp="${BUILD_SOURCEVERSIONMESSAGE/#Merge/}"
+       PR_COMMIT_ID="${tmp%%into*}"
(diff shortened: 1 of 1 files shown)


vsts.yml
```diff
@@ -27,7 +27,20 @@ phases:
 
   steps:
   - bash: |
-     echo "##vso[task.setvariable variable=gitcommit]$BUILD_SOURCEVERSION"
+     if [[ -z "${SYSTEM_PULLREQUEST_PULLREQUESTNUMBER}" ]] || "$SYSTEM_PULLREQUEST_ISFORK" == "True"; then
```
ckerr: Does the second clause here also need test brackets? e.g. 

`... || [ "${SYSTEM_PULLREQUEST_ISFORK}" == "True" ]; then`

vsts.yml
```diff
@@ -27,7 +27,20 @@ phases:
...
+       echo "##vso[task.setvariable variable=gitcommit]$BUILD_SOURCEVERSION"
+     else
+       tmp="${BUILD_SOURCEVERSIONMESSAGE/#Merge/}"
```
ckerr: TIL `${var/#Pattern/Replacement}`

vsts.yml
jkleinsc: They technically aren't needed for either one, though I guess we should be consistent.

PR #527, opened 2018-04-20
Update disabled tests

nitsakh:
Create a config file which lists the test targets with disabled tests. `run_tests` script will use this yaml file to determine and add appropriate `gtest_filter` flags while running tests


alexeykuzmin:
@nitsakh Thank you!

PR #529, opened 2018-04-23
Update run_tests to generate xml reports

nitsakh:
Add ability to generate xml report for test results.

### DIFF
--- a/script/run_tests
+++ b/script/run_tests
@@ -36,7 +36,7 @@ def GetTestsToRun(args):
         print 'Invalid test target name: ' + test_name
   return tests
 
-def RunTests(binary_tests):
+def RunTests(binary_tests, generate_report):
   if binary_tests:
     disabled_tests_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'disabled-tests.yaml')
     disabled_tests = {}
@@ -51,6 +51,9 @@ def RunTests(binary_tests):
       if disabled_tests and test_name in disabled_tests and disabled_tests[test_name]:
         exclusion_filter = '-' + ':'.join(disabled_tests[test_name])
         test_args.append('--gtest_filter=%s' % exclusion_filter)
+      # Add reporting path if enabled
+      if generate_report:
+        test_args.append("--gtest_output=xml:" + os.path.join(SOURCE_ROOT, 'test_reports', test_name + os.path.extsep + 'xml'))
       results.append((os.path.basename(test),
                       subprocess.call(test_args) == 0))
     failed = [test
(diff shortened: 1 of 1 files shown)


script/run_tests
```diff
@@ -51,6 +51,9 @@ def RunTests(binary_tests):
       if disabled_tests and test_name in disabled_tests and disabled_tests[test_name]:
         exclusion_filter = '-' + ':'.join(disabled_tests[test_name])
         test_args.append('--gtest_filter=%s' % exclusion_filter)
+      # Add reporting path if enabled
+      if generate_report:
+        test_args.append("--gtest_output=xml:" + os.path.join(SOURCE_ROOT, 'test_reports', test_name + os.path.extsep + 'xml'))
```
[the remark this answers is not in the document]
jkleinsc: @alexeykuzmin I think we can parse multiple files.  Location shouldn't matter too much.  I am going to update this PR with the appropriate CI configs.

alexeykuzmin:
```python
os.path.join(SOURCE_ROOT, 'test_reports', test_name + os.path.extsep + 'xml')
```
@nitsakh Can you please move it to a function? Looks really complicated inline )

jkleinsc:
FYI @nitsakh @alexeykuzmin:
Results in CircleCI:
https://circleci.com/gh/electron/libchromiumcontent/6405

jkleinsc:
Also FYI @nitsakh and @alexeykuzmin our Mac and Windows CIs didn't have the yaml module installed on them, so they were throwing this error:
```Traceback (most recent call last):
  File "script\run_tests", line 9, in <module>
    import yaml
ImportError: No module named yaml
```
I resolved this on Windows by running `pip install pyyaml`.  I need to do the same for our Mac build agents.

alexeykuzmin:
@jkleinsc 
We have pyyaml as a submodule.
In order to make `import yaml` work we only should do this:

```python
from config import VENDOR_DIR

PYYAML_LIB_DIR = os.path.join(VENDOR_DIR, 'pyyaml', 'lib')
sys.path.append(PYYAML_LIB_DIR)
import yaml
```

alexeykuzmin:
> https://circleci.com/gh/electron/libchromiumcontent/6405
> "2987 failures"
That's _slightly_ more than I expected ) Do all of these tests fail locally?
@nitsakh

jkleinsc:
@alexeykuzmin in regards to the failures, I think its because we are running headless and xvfb wasn't running.  I'll update to run with xvfb.

script/run_tests
nitsakh: It doesn't do much. I can change `os.path.extsep + 'xml'` to directly `.xml` is that'll make it simpler. Extension separators are same on Win, Mac and Linux anyway.

alexeykuzmin: The problem is not that it's doing to much.
It's that I have to read it through to understand what's going on there.
Having `test_args.append("--gtest_output=xml:" + get_output_file_path(test_name))` is much easier to read.

Yeah, I'm kinda nitpicking here, sorry =/

jkleinsc:
Closing in favor of #530 which continues this PR.

PR #530, opened 2018-04-25
Update run_tests to generate xml reports for CI

jkleinsc:
This is a continuation/replacement of #529.

This PR adds the necessary CI config changes to run the chromium tests and report them in the various CI platforms.

### DIFF
--- a/chromiumcontent/BUILD.gn
+++ b/chromiumcontent/BUILD.gn
@@ -357,57 +357,57 @@ if (is_electron_build && !is_component_build) {
 
     deps = [
       "//base:base_unittests",
-      "//cc:cc_unittests",
-      "//cc/blink:cc_blink_unittests",
-      "//content/test:content_browsertests",
-      "//content/test:content_unittests",
-      "//crypto:crypto_unittests",
-      "//device:device_unittests",
-      "//gin:gin_unittests",
-      "//gpu:gpu_unittests",
-      "//gpu/ipc/service:gpu_ipc_service_unittests",
-      "//ipc:ipc_tests",
-      "//media:media_unittests",
-      "//media/midi:midi_unittests",
-      "//media/mojo:media_mojo_unittests",
-      "//media/capture:capture_unittests",
-      "//mojo/common:mojo_common_unittests",
-      "//mojo/edk/system:mojo_system_unittests",
-      "//mojo/edk/test:mojo_public_bindings_unittests",
-      "//mojo/edk/test:mojo_public_system_unittests",
-      "//net:net_unittests",
-      "//ppapi:ppapi_unittests",
-      "//printing:printing_unittests",
-      "//skia:skia_unittests",
-      "//storage:storage_unittests",
-      "//sql:sql_unittests",
-      "//third_party/angle/src/tests:angle_unittests",
-      "//third_party/leveldatabase:env_chromium_unittests",
-      "//third_party/WebKit/public:test_support",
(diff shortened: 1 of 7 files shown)


jkleinsc:
Here is the test output in VSTS:
screen shot 2018-05-01 at 9 37 05 am
Test output in CircleCI can be seen here:
https://circleci.com/gh/electron/libchromiumcontent/6729?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Appveyor test output here:
https://windows-ci.electronjs.org/project/AppVeyor/libchromiumcontent/build/1.0.1464/job/1nqd1llk489c2y3k/tests

chromiumcontent/BUILD.gn
```diff
@@ -357,57 +357,57 @@ if (is_electron_build && !is_component_build) {
...
-      "//tools/gn:gn_unittests",
-      "//ui/base:ui_base_unittests",
-      #"//ui/compositor:compositor_unittests", # these are failing in chromium
-      "//ui/display:display_unittests",
-      "//ui/events:events_unittests",
-      "//ui/gl:gl_unittests",
-      "//url:url_unittests",
-      "//url/ipc:url_ipc_unittests",
-      "//v8/test/unittests:unittests",
+      # "//cc:cc_unittests",
```
ckerr: What's the plan for the rest of these tests; will these be uncommented at some point?

script/cibuild
```diff
@@ -35,25 +36,27 @@ def main():
...
-            run_ci(['-t', 'ia32'], skip_upload, use_sccache) or
-            run_ci(['-t', 'arm'], skip_upload, use_sccache) or
-            run_ci(['-t', 'arm64'], skip_upload, use_sccache))
+    return (run_ci(['-t', 'x64'], skip_upload, use_sccache, build_tests) or
+            run_ci(['-t', 'ia32'], skip_upload, use_sccache, build_tests) or
+            run_ci(['-t', 'arm'], skip_upload, use_sccache, build_tests) or
+            run_ci(['-t', 'arm64'], skip_upload, use_sccache, build_tests))
   else:
-    return run_ci([], skip_upload, use_sccache)
+    return run_ci([], skip_upload, use_sccache, build_tests)
```
ckerr: It seems like everything here is identical except for the architecture lists.

It would be slightly better to eliminate the redundancy, e.g.

```python
if 'TARGET_ARCH' in os.environ:
  archs = [os.environ['TARGET_ARCH']]
if sys.platform in ['win32', 'cygwin']:
  archs = ['x64', 'ia32']
elif sys.platform == 'linux2':
  archs = ['x64', 'ia32', 'arm', 'arm64']
else
  archs = []

for arch in archs:
  if not run_ci(['-t', arch], skip_upload, use_sccache, build_tests)
    return False

return True
```

2. ...and actually, since `run_ci()` is making assumptions on the args being passed in (it requires arg[1] to be the arch), maybe we should change `run_ci()`'s first argument from `args` to `target_arch` to remove that guesswork, and have `run_ci` build the args list itself?

chromiumcontent/BUILD.gn
jkleinsc: @ckerr this PR changes the tests to only build/run the `base:base_unittests` tests because of timeout issues building/running everything (it was timing out after running for **5** hours!).  I figure we can (in another PR) adjust the tests to pick the ones we actually want to run.

jkleinsc:
@ckerr I made the requested changes you asked for.

PR #532, opened 2018-04-29
Add patch to fix long jitter buffer delays with Opus and DTX.

ajmacd:
This backports the fix for:
https://bugs.chromium.org/p/webrtc/issues/detail?id=8488
which landed upstream in M64.

It's a critical fix for any WebRTC service that makes use of discontinuous transmission (DTX), like Slack calls. When it is triggered, the bug manifests as increased delays and severe audio dropouts.

The patch is a combination of these two CLs:
https://webrtc-review.googlesource.com/c/src/+/18181
https://webrtc-review.googlesource.com/c/src/+/44420
(the latter just to make the fix permanent)

**Testing**
I can repro the issue in Slack. I verified a build with this patch against the `electron-2-0-x` branch resolves it.

### DIFF
--- a/patches/common/webrtc/.patches.yaml
+++ b/patches/common/webrtc/.patches.yaml
@@ -21,4 +21,7 @@ patches:
   file: disable-warning-win.patch
   description:
     Disable windows warning
-
+-
+  owners: ajmacd
+  file: fix_jitter_buffer_with_opus_dtx.patch
+  description: "https://webrtc-review.googlesource.com/c/src/+/18181"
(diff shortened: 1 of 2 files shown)


ajmacd:
The field trial is complete and was deemed successful. As mentioned in the PR description (although I certainly could have made this more clear) the patch is an exact application of these two upstream CLs:
https://webrtc-review.googlesource.com/c/src/+/18181
https://webrtc-review.googlesource.com/c/src/+/44420

the latter of which reverts the field trial and makes the change permanent.

ajmacd:
Actually not an _exact_ application. I omitted the unit test revert in the latter CL, which wasn't made in the original.

ajmacd:
Sure thing! 🙌

PR #533, opened 2018-04-29
Add a patch to fix long jitter buffer delays with Opus and DTX (2.0.x edition) 

ajmacd:
This brings the patch in #532 to `electron-2-0-x`.


alexeykuzmin:
@ajmacd The patch has exactly the same changes as the upstream repo, with no modifications, right?

alexeykuzmin:
> The main discussion should probably take place at #532 

@ckerr 
Sure. But patches for different Chromium versions might be different.
Quite often we can backport a change to one branch without any trouble, but have to make some changes to backport it to another branch.

ajmacd:
@alexeykuzmin Good point. At least in this case, the changes were identical to `2-0-x` and `master`.

PR #535, opened 2018-05-01
Only build native_mksnapshot for arm and arm64

### LINKED ISSUE #509: The `native_mksnapshot` configuration should not be built on non-ARM architectures
When `script/build` is run without the `-c` parameter, it tries to build the `native_mksnapshot` configuration even on non-ARM architectures and it fails with:
```
12:45:27,570 INFO  - [exec] ninja: Entering directory `src/out-ia32/native_mksnapshot'
12:45:29,109 ERROR - [exec] ninja: error: '../../buildtools/third_party/libc++/trunk/src/algorithm.cpp', needed by 'obj/buildtools/third_party/libc++/libc++/algorithm.o', missing and no known rule to make it
```

jkleinsc:
This PR changes script/build to not build native_mksnapshot if you are not on hardware capable of building those binaries.  Fixes #509

### DIFF
--- a/script/build
+++ b/script/build
@@ -9,6 +9,7 @@ import platform
 from lib.config import MIPS64EL_GCC, set_mips64el_env, get_output_dir, \
                        SOURCE_ROOT, VENDOR_DIR, DEPOT_TOOLS_DIR, \
                        COMPONENTS
+
 from lib.ninja import run as run_ninja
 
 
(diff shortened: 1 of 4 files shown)


jkleinsc:
@alexeykuzmin good point about `create-dist`.  I updated that script as well and I also realized that `script/run-gn` could be adjusted to not run the native_mksnapshot stuff as well.

PR #538, opened 2018-05-07
Use packaged chromium source to speed up CI builds

jkleinsc:
This PR adds the capability of using a tarball cache of the chromium source code to speed up builds. 
The tarballs are stored on S3 by platform and chromium version (eg `libchromiumcontent/linux/src/63.0.3239.150/src.tar.bz2`).  If the tarball doesn't exist for the version/platform, then the source will be downloaded using `gclient sync`.  In this case if the `upload_packaged_src` argument is passed to `script/update`, then the source will be archived and uploaded to S3. 

To use this feature, pass the `use_packaged_src` argument when calling `script/update`.

The chromium source that is archived does not have the `.git` directories or the `native_client` directory.

The tarball sizes for 63.0.3239.150 are as follows:

- Linux - 1.9 gb
- Windows 1.5gb
- OSX - 1.4gb

It may be possible to further optimize these tarballs for size, but I wasn't sure what other files from chromium source we can drop.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,8 @@ jobs:
           command: script/bootstrap
       - run:
           name: Update
-          command: script/update --clean -t $TARGET_ARCH --use-bundled-sccache
+          command: script/update --clean -t $TARGET_ARCH --use-bundled-sccache --use_packaged_src --upload_packaged_src
+          no_output_timeout: 60m
       - run:
           name: Build shared library
           command: script/build -t $TARGET_ARCH -c $COMPONENT
@@ -53,7 +54,8 @@ jobs:
           command: script/bootstrap
       - run:
           name: Update
-          command: script/update --clean -t $TARGET_ARCH
+          command: script/update --clean -t $TARGET_ARCH --use_packaged_src
+          no_output_timeout: 20m
       - run:
           name: Build static library
           command: script/build -t $TARGET_ARCH -c $COMPONENT
@@ -90,7 +92,7 @@ jobs:
           command: script/bootstrap
       - run:
           name: Update
-          command: script/update --clean -t $TARGET_ARCH
+          command: script/update --clean -t $TARGET_ARCH --use-bundled-sccache --use_packaged_src
       - run:
           name: Build static library
           command: script/build -t $TARGET_ARCH -c $COMPONENT
@@ -134,7 +136,8 @@ jobs:
           command: script/bootstrap
       - run:
           name: Update
-          command: script/update --clean -t $TARGET_ARCH --use-bundled-sccache
+          command: script/update --clean -t $TARGET_ARCH --use-bundled-sccache --use_packaged_src
+          no_output_timeout: 20m
       - run:
           name: Build shared library
           command: script/build -t $TARGET_ARCH -c $COMPONENT
@@ -173,7 +176,8 @@ jobs:
           command: script/bootstrap
       - run:
           name: Update
-          command: script/update --clean -t $TARGET_ARCH
+          command: script/update --clean -t $TARGET_ARCH --use_packaged_src
+          no_output_timeout: 20m
       - run:
           name: Build static library
           command: script/build -t $TARGET_ARCH -c $COMPONENT
@@ -210,7 +214,7 @@ jobs:
           command: script/bootstrap
       - run:
           name: Update
-          command: script/update --clean -t $TARGET_ARCH
+          command: script/update --clean -t $TARGET_ARCH --use-bundled-sccache --use_packaged_src
       - run:
           name: Build static library
           command: script/build -t $TARGET_ARCH -c $COMPONENT
@@ -235,7 +239,7 @@ jobs:
       - store_test_results:
           path: test_reports
       - store_artifacts:
-          path: test_reports          
+          path: test_reports
 
   libchromiumcontent-linux-arm-shared:
     docker:
@@ -251,7 +255,8 @@ jobs:
           command: script/bootstrap
       - run:
           name: Update
-          command: script/update --clean -t $TARGET_ARCH --use-bundled-sccache
+          command: script/update --clean -t $TARGET_ARCH --use-bundled-sccache --use_packaged_src
+          no_output_timeout: 20m
       - run:
           name: Build mksnapshot
           command: script/build -t $TARGET_ARCH -c native_mksnapshot
@@ -298,7 +303,8 @@ jobs:
           command: script/bootstrap
       - run:
           name: Update
-          command: script/update --clean -t $TARGET_ARCH
+          command: script/update --clean -t $TARGET_ARCH --use_packaged_src
+          no_output_timeout: 20m
       - run:
           name: Build static library
           command: script/build -t $TARGET_ARCH -c $COMPONENT
@@ -334,7 +340,8 @@ jobs:
           command: script/bootstrap
       - run:
           name: Update
-          command: script/update --clean -t $TARGET_ARCH --use-bundled-sccache
+          command: script/update --clean -t $TARGET_ARCH --use-bundled-sccache --use_packaged_src
+          no_output_timeout: 20m
       - run:
           name: Build mksnapshot
           command: script/build -t $TARGET_ARCH -c native_mksnapshot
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -6,6 +6,7 @@ jobs:
         environment:
           TARGET_ARCH: x64
           COMPONENT: shared_library
+          CHROMIUM_BUILDTOOLS_PATH: /home/builduser/project/src/buildtools
```
nornagon: It's a bummer that we need to set this everywhere; could we perhaps convince gclient to find the root another way? Looking through [gclient_utils.py](https://chromium.googlesource.com/chromium/tools/depot_tools/+/6f64a0522bb8b6d20686028768024ebce5cfbba9/gclient_utils.py#622), it looks like it looks for a `.gclient` file and associated `.gclient_entries` (created by `gclient sync`). Is it possible the tarball is missing those files?

script/update
```diff
@@ -17,11 +18,11 @@ import lib.util as util
 from lib.config import MIPS64EL_GCC, MIPS64EL_GCC_URL, MIPS64EL_SYSROOT, \
                        set_mips64el_env, IS_ARM64_HOST, IS_ARMV7_HOST, \
                        SOURCE_ROOT, VENDOR_DIR, DEPOT_TOOLS_DIR, SRC_DIR, \
-                       COMPONENTS
-
+                       COMPONENTS,PLATFORM_KEY
```
nornagon: nit: space after `,`

script/update
```diff
@@ -38,6 +39,10 @@ solutions = [
 ]
 {cache_dir_line}
 '''
+SRC_FILES_TO_IGNORE = [
+  'src\\native_client',
+  'src/native_client'
+]
```
nornagon: Would `os.path.join(src, 'native_client')` work here?

script/update
```diff
@@ -299,6 +312,46 @@ def download(url, filename):
           break
         f.write(chunk)
 
+def get_packaged_src(clean, git_cache, nohooks, upload_src):
+  src_version = chromium_version()
+  tar_name = 'src.tar.bz2'
+  package_url = '{0}/{1}/src/{2}/{3}'.format(CHROMIUMCONTENT_SRC_URL,
+                                         PLATFORM_KEY, src_version, tar_name)
+  try:
+    print 'Trying to download ' + package_url
```
nornagon: wow it's been a while since i did python2 😅 would it make sense to use the method rather than the statement version of `print` everywhere? might make migrating to py3 easier when that day eventually comes.... any day now...

script/update
```diff
@@ -299,6 +312,46 @@ def download(url, filename):
...
+  except urllib2.URLError as err:
+    print('URLError while downloading {0}, syncing from git: {1}'.format(package_url, err.reason))
+    gclient_sync(src_version, clean, git_cache, nohooks)
+    if upload_src:
+      print 'Packaging source to ' + tar_name + '.'
+      tar_path =  os.path.join(SOURCE_ROOT, tar_name)
+      filesystem.safe_unlink(tar_path)
+      with util.scoped_cwd(SOURCE_ROOT):
+        tar_file = tarfile.open(tar_path, 'w:bz2')
+        tar_file.add('src', exclude=exclude_packaged_files)
```
nornagon: i think adding `.gclient` and `.gclient_entries` here will satisfy gclient's search for the root directory and obviate the need for `CHROMIUM_BUILDTOOLS_PATH`. Maybe. 😬

script/upload
```diff
@@ -46,10 +43,16 @@ def upload(target_arch):
   else:
     platform = PLATFORM_KEY
 
-  upload_files = glob.glob('libchromiumcontent*.tar.bz2') + glob.glob('native-mksnapshot.tar.bz2')
-  s3put(bucket, access_key, secret_key, SOURCE_ROOT,
-        'libchromiumcontent/{0}/{1}/{2}'.format(platform, target_arch, commit),
-        upload_files)
+  if args.upload_packaged_src:
+    upload_files = glob.glob('src.tar.bz2')
```
nornagon: What does `glob.glob` do here?

vsts.yml
```diff
@@ -46,18 +46,49 @@ phases:
...
     name: Bootstrap
 
   - bash: |
      set -e
      echo "===Updating for $TARGET_ARCH===" > >(tee -a buildlog.txt) 2> >(tee -a buildlog.txt >&2)
+     export CHROMIUM_BUILDTOOLS_PATH="`pwd`/src/buildtools"
+     echo "Using $CHROMIUM_BUILDTOOLS_PATH as CHROMIUM_BUILDTOOLS_PATH"
      if [ "${COMPONENT}" == "shared_library" ]; then
-       script/update --clean -t $TARGET_ARCH --use-bundled-sccache > >(tee -a buildlog.txt) 2> >(tee -a buildlog.txt >&2)
+       if [ "${TARGET_TYPE}" == "osx" ]; then
```
nornagon: Is this platform check to avoid uploading redundant tars? Perhaps add a comment?

.circleci/config.yml
jkleinsc: @nornagon I don't really think its a big deal to set the environment variable.

I did look for a .gclient file and a .gclient_entries and those files do not exist at the root of the src directory after running gclient sync.  Maybe it has to do with how we call gclient sync?

script/update
jkleinsc: Good point, I'll change it

script/upload
jkleinsc: Good point, its not really needed - I'll change it

.circleci/config.yml
nornagon: They're siblings to `src/`:

```
.
├── .gclient
├── .gclient_entries
└── src
```

script/update
```diff
@@ -300,6 +317,46 @@ def download(url, filename):
...
+    raise
+
+
+def exclude_packaged_files(filename):
+  if ".git" in filename:
+    return True
+  else:
+    should_exclude = filename in SRC_FILES_TO_IGNORE
+    if not should_exclude:
+      print('Adding {0} to tarfile'.format(filename))
```
nornagon: Is this debugging output that's no longer necessary? Seems like this would print a LOT

vsts.yml
```diff
@@ -71,7 +103,7 @@ phases:
      if [ "${COMPONENT}" == "shared_library" ]; then
        script/sccache --stop-server > >(tee -a buildlog.txt) 2> >(tee -a buildlog.txt >&2)
      else
-       if [ "${BUILD_REASON}" == "Schedule" ]; then
+       if [ "Schedule" == "Schedule" ]; then
```
nornagon: ...?

jkleinsc: @nornagon always true! 😄 I'm testing if our libcc tests still run/build properly.  This is a temporary change

script/update
jkleinsc: CircleCI has this thing where if it doesn't receive output in a certain amount of time, it will cancel the job.  I added this output here to avoid that.  I can probably remove it now that I know approximately how long it will take to run that bit so I can set an appropriate timeout

nornagon: Ah cool, sounds good!

zcbenz:
Just for your information, we were doing this before and I had a service providing universal source tarballs https://github.com/zcbenz/chromium-source-tarball.

I dont remember the reason and when we switched away from it (I was not making the call), but if you are interested in it, I can bring my service back to life.

jkleinsc:
@zcbenz would your solution have any advantage over this one?  It seems to me that using S3 means we don’t need to run another service.

zcbenz:
It hosts assets on GitHub Releases so it only needs a $5 cloud machine to run. It does not make much difference for us since the bill is on GitHub anyway.

script/lib/filesystem.py
```diff
@@ -36,3 +41,19 @@ def safe_unlink(path):
   except OSError as e:
     if e.errno != errno.ENOENT:
       raise
+
+def download_and_extract(destination, url):
+  print url
```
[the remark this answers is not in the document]
jkleinsc: @nornagon this function (and print statement) isn’t new, I just moved it for common use.  It is used when bootstrapping Electron in order to down load libcc.

PR #539, opened 2018-05-08
Don't lock the scheme registry

nornagon:
This effectively just disables some DCHECKs.


nornagon:
Abandoning because this is already disabled by dcheck.patch when `is_electron_build = true`

PR #540, opened 2018-05-10
Make power_save_blocker visible from //electron:*

nornagon:


alexeykuzmin:
@nornagon You have to add the new patch to `patches/common/chromium/.patches.yaml`.

nornagon:
It's not a new patch.

PR #543, opened 2018-05-14
FIX create-dist uses libchromiumcontent/chromiumcontent

gavignus:
Fix create-dist script by pointing it to the right directory with static libraries


alexeykuzmin:
cc @nornagon

PR #550, opened 2018-05-21
Add patch to fix remote debugging in 'Select Element Button'

codebytere:
Add patch to fix [this](https://bugs.chromium.org/p/chromium/issues/detail?id=837145) bug, which will theoretically fix the current failing `<webview> loads devtools extensions registered on the parent window` test for Chromium 66.

Patch pulled from CEF [here](https://bitbucket.org/chromiumembedded/cef/commits/6c5418d).

/cc @alexeykuzmin @deepak1556


alexeykuzmin:
So it's basically a backport of https://chromium-review.googlesource.com/c/chromium/src/+/953144?
Can you please then rename a file accordingly and add it to a `.patches.yaml`?

alexeykuzmin:
The patch fails to apply =(

PR #551, opened 2018-05-22
[WIP] Add patch to fix some guestinstance tests

codebytere:
After enabling verbose logging, it appeared that some guestinstance tests were crashing as a result of 
```
[25755:0522/121849.511223:ERROR:sync_point_manager.cc(136)] Client did not release sync token as expected
```
errors. 

This error was also seen [here](https://bitbucket.org/chromiumembedded/cef/issues/2434/client-did-not-release-sync-token-as), and subsequently repaired in chromium with [this CL](https://chromium-review.googlesource.com/c/chromium/src/+/969639), which is the patch i've backported in this PR.

/cc @alexeykuzmin @deepak1556

### DIFF
--- a/patches/common/chromium/.patches.yaml
+++ b/patches/common/chromium/.patches.yaml
@@ -275,3 +275,12 @@ patches:
     configurations of the inspector app. See
     https://chromium-review.googlesource.com/c/chromium/src/+/953144 for more
     details.
+-
+  owners: codebytere
+  file: backport_3743d0f.patch
+  description: |
+    Fixes some failing guestinstance attribute tests by removing the DCHECK for 
+    wait order number and the spammy invalidclient wait logging. See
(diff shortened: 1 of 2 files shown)


alexeykuzmin:
`> backport_3743d0f.patch failed to apply` =(

deepak1556 (changes requested):
https://chromium-review.googlesource.com/c/chromium/src/+/969968 should also be backported.

alexeykuzmin:
There is no need to backport commit [`3743d0f`](https://chromium.googlesource.com/chromium/src/+/3743d0f5f0614909a354a8df3af879f0eff21c42), it has already been backported to M66 in commit [`4ec34bb`](https://chromium.googlesource.com/chromium/src/+/4ec34bb46504fcfce52bb6644b570cd48157a79f).

alexeykuzmin:
> https://chromium-review.googlesource.com/c/chromium/src/+/969968 should also be backported.

I'll do it separately.

PR #564, opened 2018-06-19
Revert "backport: Allow lookup of videodecoder service to fix HW video decoding on macOS 10.13"

alexeykuzmin:
Reverts electron/libchromiumcontent#558


alexeykuzmin:
The patch is fixed in #565, there's no need to revert the original change. Closed.

PR #570, opened 2018-06-20
Make v8 api forwards compatible with v8 6.7

nornagon:
Cribbed from https://github.com/nodejs/node/commit/2a3f8c3a83e3820d3c601bdf02eda8db462027aa

### DIFF
--- a/patches/common/v8/.patches.yaml
+++ b/patches/common/v8/.patches.yaml
@@ -60,3 +60,9 @@ patches:
   file: backport_2eb23a17.patch
   description: |
     Part (2/2) of backporting the array splice perf regression caused in ch66
+-
+  owners: nornagon
+  file: node_2a3f8c3a8.patch
+  description: |
+    Patch lifted from node's patches to their vendored copy of v8.
+    > patch the V8 API to be forward compatible with 6.7
--- a/patches/common/v8/node_2a3f8c3a8.patch
+++ b/patches/common/v8/node_2a3f8c3a8.patch
@@ -1,0 +1,549 @@
+From 2a3f8c3a83e3820d3c601bdf02eda8db462027aa Mon Sep 17 00:00:00 2001
+From: Peter Marshall <p.s.marshall0@gmail.com>
+Date: Thu, 12 Apr 2018 16:42:10 +0200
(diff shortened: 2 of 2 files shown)


alexeykuzmin:
Why do we need this?

nornagon:
Building 3rd-party modules against the version of node we depend on (v10.2.0) results in modules that are incompatible with the version of v8 in chrome 66 (v6.6.346.32), because node v10.2.0 includes this patch in its vendored v8.

In particular, the thing i ran into is that node v10.2.0 has this signature for `v8::FunctionTemplate::New`:
```cpp
Local<FunctionTemplate> FunctionTemplate::New(
    Isolate* isolate, FunctionCallback callback, v8::Local<Value> data,
    v8::Local<Signature> signature, int length, ConstructorBehavior behavior,
    SideEffectType side_effect_type);
```

but the function signature for `v8::FunctionTemplate::New` in chromium's version of v8 (v6.6.346.32) is:
```cpp
Local<FunctionTemplate> FunctionTemplate::New(
    Isolate* isolate, FunctionCallback callback, v8::Local<Value> data,
    v8::Local<Signature> signature, int length, ConstructorBehavior behavior);
```

So a module built against node v10.2.0's v8 headers won't work with our v8.

alexeykuzmin:
> So a module built against node v10.2.0's v8 headers won't work with our v8.

Is it expected to do so?
@zcbenz ,can you help us to understand how the thing should work here?

nornagon:
ah, i see that the `create-node-headers.py` script in e/e overrides nodejs's v8 headers with the headers from chrome. I should probably figure out a way to do that in the gn build.

MarshallOfSound:
In the past we've tried to make this work (building native modules for electron by building them on the right node version) but afaik it has _sometimes_ worked on *nix but it has never worked on windows.

We have `electron-rebuild` and friends to make getting around this problem easy

nornagon:
`electron-rebuild` doesn't work for development tho (i.e. developing on an unreleased electron)

MarshallOfSound:
This is correct but there are helper scripts in the electron repo to rebuild native modules with electron generated headers.

PR #572, opened 2018-06-21
Revert "Disable a DCHECK in process_metrics_win.cc"

torycl:
Revert "Disable a DCHECK in process_metrics_win.cc"
This reverts commit b9a68acb7e1f93193540cdf94f73ad6ba22c7900.

It seems it no longer cause the problem reported in Electron unit test named 
"app module getAppMetrics() API returns memory and cpu stats of all running electron processes".


torycl:
I will push the branch again and recreate the pull request

PR #573, opened 2018-06-21
Revert "Disable a DCHECK in process_metrics_win.cc"

torycl:
This reverts commit b9a68acb7e1f93193540cdf94f73ad6ba22c7900.

It seems this DCHECK is ok and should be reenabled now.
Electron doesn't crash with this DCHECK being enabled: https://windows-ci.electronjs.org/project/AppVeyor/electron/build/1.0.5964

Fixes electron/electron#13317.

### DIFF
--- a/patches/common/chromium/dcheck.patch
+++ b/patches/common/chromium/dcheck.patch
@@ -237,21 +237,3 @@ index 674b0e9a909c..a1bff6e40f56 100644
     if (base_computed_style_ && computed_style)
       DCHECK(*base_computed_style_ == *computed_style);
   #endif
-diff --git a/base/process/process_metrics_win.cc b/base/process/process_metrics_win.cc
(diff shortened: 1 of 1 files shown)


alexeykuzmin:
@torycl Can you please create a PR in `electron/electron` with only libcc ref updated to run Electron tests with that DCHECK being enabled?
Only simply update https://github.com/electron/electron/pull/13361 you already have.

alexeykuzmin:
Electron builds with this DCHECK do not crash:
https://windows-ci.electronjs.org/project/AppVeyor/electron/build/1.0.5964

alexeykuzmin:
@zcbenz Must be not stable then, I've run Electron builds with this change several times and didn't see a single crash.
We'll revert it back for now then. Thank you!

PR #575, opened 2018-06-25
[WIP] build: move printing support behind feature flag

deepak1556:
Ref https://github.com/electron/maintainers/issues/174

- [ ] MacOS
- [ ] Linux
- [ ] Windows


chromiumcontent/BUILD.gn
```diff
@@ -465,15 +427,12 @@ static_library("pepper_flash") {
     "//chrome/renderer/pepper/pepper_shared_memory_message_filter.cc",
     "//chrome/renderer/pepper/pepper_shared_memory_message_filter.h",
   ]
-  deps = [
-    "//media:media_features",
-    "//ppapi/host",
-    "//ppapi/proxy",
-    "//ppapi/proxy:ipc",
-    "//ppapi/shared_impl",
+  data_deps = [
```
nornagon: it's surprising to me that mojom and flapper turn into data_deps here. can you explain why this change is needed?

chromiumcontent/BUILD.gn
```diff
@@ -110,41 +108,11 @@ if (is_electron_build && is_component_build) {
     public_deps = [ "//services/device:lib" ]
   }
 
-  static_library("fx_agg") {
```
nornagon: how come we no longer need this?

deepak1556: We do need it, but its already built by pdfium as a dependency https://github.com/electron/libchromiumcontent/pull/575/files#diff-6bcdc4060f1a8322ee9b67ee3fcc7078R105, this step is not needed.

chromiumcontent/BUILD.gn
deepak1556: No specific reason, I was just testing pepper flash without linking those. Sorry for the mislead here.

chromiumcontent/BUILD.gn
nornagon: 👍

deepak1556:
Will be moving the feature behind flag in the Electron GN build config.

PR #577, opened 2018-06-27
Do not check for unique origin in CacheStorage

zcbenz:
In Electron we may have scripts running without an origin (for example from Node.js), and it will trigger an assertion in `CacheStorage`. The chromium commit introduced the check is https://chromium.googlesource.com/chromium/src/+/b6a21896d9687ff8c7c6aa809e45b3efe2fefb50%5E%21/#F0.

This patch intends to disable a behavior instead of temporarily disabling a check, so I did not put it in `dcheck.patch`.

This should be able to fix the second crash in https://github.com/electron/electron/issues/12835#issuecomment-398941227.

### DIFF
--- a/patches/common/chromium/.patches.yaml
+++ b/patches/common/chromium/.patches.yaml
@@ -306,3 +306,9 @@ patches:
   description: |
     Fixes a bunch of warnings/errors in WTL code that isn't used by Chromium,
     but is used by Electron.
+-
+  owners: zcbenz
+  file: no_cache_storage_check.patch
+  description: |
+    Do not check for unique origin in CacheStorage, in Electron we may have
+    scripts running without an origin.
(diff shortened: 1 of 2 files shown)


deepak1556:
@zcbenz we already disable most of the storages for unique origins in [content_settings_observer](https://github.com/electron/electron/blob/master/atom/renderer/content_settings_observer.cc). There is no hook for cache storage, but it already has unique origin checks which disables all its operations https://cs.chromium.org/chromium/src/content/browser/cache_storage/cache_storage_dispatcher_host.cc?g=0&l=41 ,  however there are some unsolicited calls from the quota manager which doesn't have origin checks that caused this crash https://cs.chromium.org/chromium/src/content/browser/cache_storage/cache_storage_quota_client.cc. The only way to fix it with minimal patch is to disable quota manager for unique origins, since we already disable storage operations I think it should be fine disabling the manager https://cs.chromium.org/chromium/src/content/browser/storage_partition_impl.cc?g=0&l=574. Thoughts ?

zcbenz:
> The only way to fix it with minimal patch is to disable quota manager for unique origins, since we already disable storage operations I think it should be fine disabling the manager https://cs.chromium.org/chromium/src/content/browser/storage_partition_impl.cc?g=0&l=574. Thoughts ?

@deepak1556 It looks good to me to disable the quota manager.

PR #580, opened 2018-06-29
Remove some new private APIs in MAS build

zcbenz:
Refs https://github.com/electron/electron/issues/13410.

### DIFF
--- a/patches/mas/chromium/no_private_api.patch
+++ b/patches/mas/chromium/no_private_api.patch
@@ -228,6 +228,34 @@ index 6b955bdb3724..1775f3b3c1d4 100644
  }
  
  // You are about to read a pretty disgusting hack. In a static initializer,
+diff --git a/media/audio/BUILD.gn b/media/audio/BUILD.gn
+index 10d786f..2d96723 100644
+--- a/media/audio/BUILD.gn
++++ b/media/audio/BUILD.gn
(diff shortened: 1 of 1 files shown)


patches/mas/chromium/no_private_api.patch
```diff
@@ -284,6 +356,62 @@ index dfba0bded9e3..876f96999f53 100644
...
+ namespace sandbox {
+ 
+@@ -46,7 +48,11 @@ std::unique_ptr<SeatbeltExtension> SeatbeltExtension::FromToken(
+ 
+ bool SeatbeltExtension::Consume() {
+   DCHECK(!token_.empty());
++#ifndef MAS_BUILD
+   handle_ = sandbox_extension_consume(token_.c_str());
++#else
++  handle_ = -1;
```
deepak1556: Should this be `handle_ = 0` because of the DCHECK in the dtor https://cs.chromium.org/chromium/src/sandbox/mac/seatbelt_extension.cc?sq=package:chromium&g=0&l=25 ?

zcbenz: The MAS build is not supposed to invoke any sandbox code, so it do not make a difference here.

But triggering the DCHECK is probably better since it would warn we have run into code path that should not be invoked.

PR #584, opened 2018-07-04
build: improve the download script to show percentage / ratio progress of libcc download

MarshallOfSound:
The easiest way to show how much better this is than the old dots is a GIF

https://i.imgur.com/6214pKI.gif

cc @deepak1556 I did it 😄

### DIFF
--- a/script/lib/filesystem.py
+++ b/script/lib/filesystem.py
@@ -42,15 +42,24 @@ def safe_unlink(path):
     if e.errno != errno.ENOENT:
       raise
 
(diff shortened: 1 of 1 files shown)


zcbenz:
When using a light theme, the first number is invisible 😄 
screen shot 2018-07-10 at 17 29 32

MarshallOfSound:
@zcbenz Maybe we should change it to `-> XX%` 🤔 I'll put that up in PR 👍

PR #585, opened 2018-07-05
Add appveyor override

jkleinsc:
Our Appveyor instance was recently changed to by default use a VM that only contains VS2017 for libcc builds.  That's fine for master and 3-0-x, but older versions need VS2017, so this PR is an Appveyor override to use VS2015 instead of the default of VS2017


jkleinsc:
It appears this isn't needed as we use the same Windows VM for 2-0-x and newer

PR #589, opened 2018-07-10
Disable DCHECKs related to ScrollBegin event

zcbenz:
When testing https://github.com/electron/electron/issues/13137 I have encountered these assertions. I grouped them together since they are all related to the ScrollBegin event.


alexeykuzmin:
Any idea why we hit it?

zcbenz:
> Any idea why we hit it?

I think it is related to our code  in `NativeWindowMac` listening to the `NSScrollWheel` events.

PR #590, opened 2018-07-10
build: improve percentage output in download script

MarshallOfSound:
Follow up to #584

### DIFF
--- a/script/lib/filesystem.py
+++ b/script/lib/filesystem.py
@@ -45,7 +45,7 @@ def safe_unlink(path):
 def byte_to_mb(n):
   return str(n / 1024 / 1024) + 'MB'
 
-def download_and_extract(destination, url):
+def download_and_extract(destination, url, verbose):
   print url
   with tempfile.TemporaryFile() as t:
     with contextlib.closing(urllib2.urlopen(url)) as u:
@@ -57,14 +57,16 @@ def download_and_extract(destination, url):
         done += len(chunk)
         if not len(chunk):
           break
-        percent = '{0:.2f}%'.format(round(float(done) / float(total), 4) * 100)
-        ratio = '(' + byte_to_mb(done) + '/' + byte_to_mb(total) + ')'
-        line = '-> ' + percent + ' ' + ratio
-        sys.stderr.write(line.ljust(last_length) + '\r')
-        last_length = len(line)
-        sys.stderr.flush()
+        if verbose:
+          percent = '{0:.2f}%'.format(round(float(done) / float(total), 4) * 100)
+          ratio = '(' + byte_to_mb(done) + '/' + byte_to_mb(total) + ')'
+          line = '-> ' + percent + ' ' + ratio
+          sys.stderr.write(line.ljust(last_length) + '\r')
+          last_length = len(line)
+          sys.stderr.flush()
         t.write(chunk)
-    sys.stderr.write('\nExtracting...\n')
-    sys.stderr.flush()
+    if verbose:
+      sys.stderr.write('\nExtracting...\n')
+      sys.stderr.flush()
     with tarfile.open(fileobj=t, mode='r:bz2') as z:
       z.extractall(destination)
--- a/script/lib/filesystem.py
+++ b/script/lib/filesystem.py
@@ -57,12 +57,11 @@ def download_and_extract(destination, url):
         done += len(chunk)
         if not len(chunk):
           break
-        percent = '{0:.2f}'.format(round(float(done) / float(total), 4) * 100) + '%'
+        percent = '{0:.2f}%%'.format(round(float(done) / float(total), 4) * 100)
         ratio = '(' + byte_to_mb(done) + '/' + byte_to_mb(total) + ')'
         line = '-> ' + percent + ' ' + ratio
-        c_length = last_length
-        last_length = len(line)
         sys.stderr.write(line.ljust(c_length) + '\r')
+        last_length = len(line)
         sys.stderr.flush()
         t.write(chunk)
     sys.stderr.write('\nExtracting...\n')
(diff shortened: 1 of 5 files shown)


script/lib/filesystem.py
```diff
@@ -51,15 +51,18 @@ def download_and_extract(destination, url):
...
       done = 0
+      last_length = 0
       while True:
         chunk = u.read(1024)
         done += len(chunk)
         if not len(chunk):
           break
-        sys.stderr.write('                                   \r')
-        percent = str(round(float(done) / float(total), 4) * 100) + '%'
+        percent = '{0:.2f}'.format(round(float(done) / float(total), 4) * 100) + '%'
```
ckerr: (minor) I think you could avoid joining two strings by making the first string `'{0:.2f}%%'`

script/lib/filesystem.py
```diff
@@ -51,15 +51,18 @@ def download_and_extract(destination, url):
...
         ratio = '(' + byte_to_mb(done) + '/' + byte_to_mb(total) + ')'
-        sys.stderr.write(percent + ' ' + ratio + '\r')
+        line = '-> ' + percent + ' ' + ratio
+        c_length = last_length
+        last_length = len(line)
+        sys.stderr.write(line.ljust(c_length) + '\r')
```
ckerr: `c_length` is unnecessary if you swap the write and calculate lines:

```python
line = ...
sys.stderr.write(line.ljust(last_length) + '\r')
last_length = len(line)
```

MarshallOfSound: I felt there would be a way to do that.... Mind blanked at midnight 😆

script/lib/filesystem.py
```diff
@@ -51,15 +51,17 @@ def download_and_extract(destination, url):
...
         done += len(chunk)
         if not len(chunk):
           break
-        sys.stderr.write('                                   \r')
-        percent = str(round(float(done) / float(total), 4) * 100) + '%'
+        percent = '{0:.2f}%%'.format(round(float(done) / float(total), 4) * 100)
         ratio = '(' + byte_to_mb(done) + '/' + byte_to_mb(total) + ')'
-        sys.stderr.write(percent + ' ' + ratio + '\r')
+        line = '-> ' + percent + ' ' + ratio
+        sys.stderr.write(line.ljust(c_length) + '\r')
```
jkleinsc: @MarshallOfSound should this be `last_length` instead of `c_length`?  CI is failing with:
```
Traceback (most recent call last):
  File "script/update", line 367, in <module>
    sys.exit(main())
  File "script/update", line 79, in main
    get_packaged_src(args.clean, git_cache, nohooks, args.upload_packaged_src)
  File "script/update", line 331, in get_packaged_src
    filesystem.download_and_extract(SOURCE_ROOT, package_url)
  File "/home/builduser/project/script/lib/filesystem.py", line 63, in download_and_extract
    sys.stderr.write(line.ljust(c_length) + '\r')
NameError: global name 'c_length' is not defined
```

script/lib/filesystem.py
```diff
@@ -51,15 +51,17 @@ def download_and_extract(destination, url):
...
       done = 0
+      last_length = 0
       while True:
         chunk = u.read(1024)
         done += len(chunk)
         if not len(chunk):
           break
-        sys.stderr.write('                                   \r')
-        percent = str(round(float(done) / float(total), 4) * 100) + '%'
+        percent = '{0:.2f}%%'.format(round(float(done) / float(total), 4) * 100)
```
jkleinsc: @MarshallOfSound I just realized there is an extra `%` here, which ends up producing output like this:
```
-> 0.10%% (1MB/1729MB)
```
(From https://circleci.com/gh/electron/libchromiumcontent/8785?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

script/lib/filesystem.py
```diff
@@ -51,15 +51,17 @@ def download_and_extract(destination, url):
     with contextlib.closing(urllib2.urlopen(url)) as u:
       total = int(u.headers['content-length'])
       done = 0
+      last_length = 0
       while True:
         chunk = u.read(1024)
```
jkleinsc: @MarshallOfSound I think this change introduced in  #584 is making our downloads really slow on our Mac CI.  Before #584 this was  `chunk = u.read(1024*1024)`, so now we are effectively reading smaller chunks and the effect is that downloads of libcc are taking over an hour our nightly builds, eg see: https://github.visualstudio.com/electron/_build/results?buildId=2522&view=logs (bootstrap times out job at an hour) vs https://github.visualstudio.com/electron/_build/results?buildId=2169 where bootstrap took 9 minutes

MarshallOfSound: Hm, I did that to cosmetically improve the download number incrementing.  I guess if writing to stdout is slow on CI it could have a very noticeable impact.  Tbh if this was written in node we could have streams and do this a lot nicer.  For now let's just bump it back up to 1024*1024

jkleinsc: I'm testing it now with 1mb (1024*1024) vs 1k.

jkleinsc: @MarshallOfSound yeah... its much faster now!
This was with 1024: https://github.visualstudio.com/electron/_build/results?buildId=2380&view=logs (1 hour to pull from S3)
This was with 1024*1024: 
https://github.visualstudio.com/electron/_build/results?buildId=2534&view=logs (13 minutes to pull from S3).

I believe AppVeyor was having the same issue, but strangely CircleCI was not.  Maybe because CircleCI uses AWS vms.

MarshallOfSound: Here's a fun question, how fast is it if you remove all the logging during download on CI?  Could we save another few minutes turning off these logs during CI builds?

jkleinsc: Maybe... I can check

jkleinsc: Removing the logging reduced the update step by 5 minutes.

I moved it to a verbose setting which is off by default.

PR #591, opened 2018-07-10
fix: backport of process metrics fix from chromium

torycl:
Backport requested in and
fixes https://github.com/electron/electron/issues/13605

PR created to check if it will start to work on CI machines. It always worked on local machines.


torycl:
I have moved the diff from dchecks.patch to separate file. The function has moved in the chromium repo, so I have also fixed line numbers so it fits to out current Chromium version.

patches/common/chromium/.patches.yaml
```diff
@@ -322,3 +322,8 @@ patches:
     https://chromium.googlesource.com/chromium/src/+/f47b361887a31cccf42a6e21a82bccf28372bdaa%5E%21
     In the long term we should investigate why it happened, and take a more
     formal fix. But for now I'm just make this assertion silently pass away.
+
+  owners: torycl
```
alexeykuzmin: There should be `- ` on the previous line to indicate a new list item:
```yaml
-
  owners: torycl
```

patches/common/chromium/.patches.yaml
```diff
@@ -322,3 +322,8 @@ patches:
...
+  file: backport_e3a883075.patch
+  description: |
+    https://chromium.googlesource.com/chromium/src/+/e3a883075f699e67ab47d9991ac55b143017cfc3
```
alexeykuzmin: Please add a line saying the this change has landed in version `67.0.3380.0`.
It will make it easier to make a decision if it's still needed at the start of the next Chromium upgrade.

patches/common/chromium/.patches.yaml
torycl: Done

patches/common/chromium/.patches.yaml
torycl: Done

torycl: Thank you for review

alexeykuzmin:
@torycl ^

alexeykuzmin:
> Trop won't work on unmerged PR's

yeah, I should get more sleep I guess =/
@MarshallOfSound

zcbenz:
@torycl The crash still happens when running tests on Windows, currently the CI on master branch is failing after including this change:

```
python : [3224:0718/020420.879:FATAL:process_metrics_win.cc(332)] Check failed: result. : The handle is invalid. (0x6)
```

PR #595, opened 2018-07-11
build: ship mojo v8 proxy resolver

deepak1556:


alexeykuzmin:
@jkleinsc It looks like we have a Windows CI machines misconfiguration here.

> Exception: Visual Studio Version 2015 (from GYP_MSVS_VERSION) not found.

PR #605, opened 2018-07-19
build: refactor mas patches

nornagon:
This PR merges the MAS patches into the common/ directory, so they'll be applied unconditionally at patch time, and their functionality toggled at build time. This is because the GN build should not rely on the `MAS_BUILD` environment variable, but instead only on the GN args system (e.g. so it's possible to have `out/Release` and `out/Release-mas` in parallel, without having to coordinate what environment variables are set).


jkleinsc (approved):
LGTM

PR #606, opened 2018-07-19
build: add --folder arg to apply-patches

nornagon:
This allows one to apply folders of patches by hand more easily.


alexeykuzmin:
@nornagon 
It is definitely not obvious, but you could just use `./script/patch.py -d path/to/a/folder`.

nornagon:
ahh, totally missed that! thanks. I'll remove this feature and add a note pointing to patch.py.

PR #607, opened 2018-07-20
chore: Chromium 67

alexeykuzmin:
Project Board https://github.com/electron/electron/projects/11
PR to Electron https://github.com/electron/electron/pull/13756


patches/common/chromium/.patches.yaml
```diff
@@ -311,3 +311,9 @@ patches:
   description: |
     Make chrome's install-sysroot scripts point to our custom sysroot builds,
     which include extra deps that Electron needs (e.g. libnotify)
+-
+  owners: marshallofsound
+  file: string_piece_check_disable.patch
+  description: |
+    Disables some static checks on constexpr's in the string_piece helper to
+    avoid incompatibilities with node's CHECK macro
```
alexeykuzmin: @MarshallOfSound 
Please preserve newlines at the end of files.

patches/common/chromium/string_piece_check_disable.patch
```diff
@@ -0,0 +1,28 @@
...
+   constexpr void remove_prefix(size_type n) {
+-    CHECK(n <= length_);
++    // CHECK(n <= length_);
+     ptr_ += n;
+     length_ -= n;
+   }
+
+   constexpr void remove_suffix(size_type n) {
+-    CHECK(n <= length_);
++    // CHECK(n <= length_);
```
alexeykuzmin: @MarshallOfSound Looks scary.
https://github.com/electron/electron/projects/11#card-11512594

deepak1556: The patch is not needed, the fix is on electron side.

alexeykuzmin:
@MarshallOfSound @deepak1556 
I'll clean up the branch later this week to remove all redundant commits.

patches/common/chromium/string_piece_check_disable.patch
MarshallOfSound: Heh, heavy-handed approach 😆

chromiumcontent/build_libs.py
```diff
@@ -143,11 +143,12 @@ def gen_list(out, name, obj_dirs, exclude_dirs=[]):
         out,
         "obj_components",
         [
-            "components/appdu",
```
MarshallOfSound: Oh my.... I spent like a solid hour this afternoon wondering why this didn't work.... 😢

chromiumcontent/args/ffmpeg.gn
```diff
@@ -3,11 +3,15 @@ is_component_build = false
...
 enable_nacl = false
-enable_widevine = true
 proprietary_codecs = false
 is_component_ffmpeg = true
 ffmpeg_branding = "Chromium"
 
+# As of Chromium 67.0.3396.99 widevine is only supported on Mac and Windows.
+if (target_os != "linux") {
+  enable_widevine = true
+}
```
nornagon: This seems like a bummer; is upstream no longer supporting this?

deepak1556: We don't have the chrome component updater setup in electron, also for proprietary reasons we ask users to get their own widevine libraries, electron only integrates mechanism required to load cdm. So this build flag will not affect us directly.

alexeykuzmin: @nornagon It is not supported on Linux in Ch67. I have no idea if there are plans to support it in the future.

deepak1556 (approved):
LGTM

PR #620, opened 2018-07-27
fix: [gn] port crashpad patch from fork to patch

nornagon:
Since the GN build uses the crashpad that's referenced by Chromium's
DEPS, port over the one semantic change from Electron's crashpad fork to
be a patch.

### DIFF
--- a/patches/common/crashpad/.patches.yaml
+++ b/patches/common/crashpad/.patches.yaml
@@ -1,0 +1,8 @@
+repo: src
+patches:
+-
+  owners: nornagon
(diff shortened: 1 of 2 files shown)


MarshallOfSound:
I can't rerun the VSTS builds that broke, can you push up again to kick the builds 👍

MarshallOfSound:
Also worth noting is this a thing that we could potentially upstream?

nornagon:
Patch successfully upstreamed. Can be removed when we pass M71.

PR #621, opened 2018-07-27
chore: refactor script/sccache

alexeykuzmin:
Handle only custom arguments, pass all others to the `sccache` binary.

/cc @nornagon

### DIFF
--- a/script/sccache
+++ b/script/sccache
@@ -6,30 +6,28 @@ import sys
 
 import lib.sccache as sccache
 
+
 def parse_args():
   parser = argparse.ArgumentParser(description='Run sccache`')
+
   parser.add_argument('--azure_container', help='Name of azure container to use')
   parser.add_argument('--azure_connection', help='Name of azure connection to use')
-  parser.add_argument('-s', '--show-stats', action='store_true', help='Show cache statistics')
-  parser.add_argument('--start-server', action='store_true', help='start background server')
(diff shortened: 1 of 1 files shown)


script/sccache
```diff
@@ -6,30 +6,28 @@ import sys
 
 import lib.sccache as sccache
 
+
 def parse_args():
   parser = argparse.ArgumentParser(description='Run sccache`')
+
   parser.add_argument('--azure_container', help='Name of azure container to use')
   parser.add_argument('--azure_connection', help='Name of azure connection to use')
```
nornagon: Can we get rid of all argument parsing and just pass the args straight through to sccache? I don't think these args are necessary, all they do is set an environment variable.

alexeykuzmin: @jkleinsc ^

jkleinsc: @alexeykuzmin @nornagon those args are needed for VSTS because they are stored in secret variables that can't be accessed directly as environment variables but they can be passed along in a command invocation, but sccache doesn't support them that way.

nornagon: Can they not be passed like:

```sh
AZURE_CONTAINER=$azure_container_secret_variable ninja out/Default ...
```
?

jkleinsc: Unfortunately, no: https://docs.microsoft.com/en-us/vsts/pipelines/process/variables?view=vsts&tabs=designer%2Cbatch#set-in-script

PR #623, opened 2018-07-31
Backport "Enforce flush ordering between contexts in GPU process on macOS."

jlfwong:
This is a backport of https://chromium-review.googlesource.com/c/chromium/src/+/687832 targeting the 2-0-x branch of Electron.

This is a crucial patch for Figma, since without it mac users will see flickering when their GPU is under heavy load.

@poiru


poiru:
@jlfwong Thanks! Seems like the patch failed to apply so I opened a fixed PR in #623 instead to get the builds going sooner.

PR #625, opened 2018-07-31
fix: implement ArrayBuffer Realloc

codebytere:
This PR is done in concert with [this commit](https://github.com/electron/node/commit/6ef4b2a3422d40dad6c96d8e2899136146bec866) in the `v10.6.0` Node upgrade for Chromium 67 to allow for overwriting of ReallocateBufferMemory to use v8's allocator.


alexeykuzmin:
> fatal: corrupt patch at line 33
add_realloc.patch failed to apply

alexeykuzmin:
Is it possible to upstream this change?

codebytere:
@alexeykuzmin good q, i'll try to find out by asking a v8 person.
i think i left off a newline, will fix

patches/common/v8/.patches.yaml
```diff
@@ -1,5 +1,16 @@
...
 patches:
+-
+  owners: codebytere
+  file: add_realloc.patch
+  description: |
+    Blink overrides ArrayBuffer's allocator with its own one, while Node simply
+    uses malloc and free, so we need to use v8's allocator in Node. As part of the
+    10.6.0 upgrade, we needed to make SerializerDelegate accept an allocator
+    argument in its constructor, and override ReallocateBufferMemory and
+    FreeBufferMemory to use the allocator. We cannoy simply allocate and then memcpy
```
ckerr: s/cannoy/cannot/

patches/common/v8/.patches.yaml
```diff
@@ -1,5 +1,16 @@
...
+    when we override ReallocateBufferMemory, so we therefore need to implement
```
ckerr: Why not? That is, why not implement `ReallocBufferMemory` in terms of 
existing methods `AllocateUninitialized` and `Free`?

```cpp
ReallocateBufferMemory(src, size) {
  void * tgt = AllocateUninitialized(size);
  if (tgt != NULL) {
    memcpy(tgt, src, size);
    Free(src);
  }
  return tgt;
}

File: patches/common/v8/add_realloc.patch [hunk 3]
```diff
@@ -0,0 +1,34 @@
...
+   }
+
++  void* Realloc(void* data, size_t length) override {
++    return realloc(data, length);
++  }
++
+   void Free(void* data, size_t) override { free(data); }
+ };
+
+
```
alexeykuzmin: The file should have end with a new line.

patches/common/v8/.patches.yaml
codebytere: @ckerr:
`ReallocateBufferMemory` has the following signature:
```void* ReallocateBufferMemory(void* old_buffer,
                               size_t size,
                               size_t* actual_size)
```
and we don't have access to the old buffer's size in order to use `memcpy` as it is plain memory

alexeykuzmin:
The builds still fail trying to apply the patch.
I don't want to approve something that is bound to change )

alexeykuzmin:
```
error: patch failed: src/v8/include/v8.h:4318
error: src/v8/include/v8.h: patch does not apply
add_realloc.patch failed to apply
```

alexeykuzmin:
@codebytere Build still fail =(

patches/common/chromium/add_realloc.patch
```diff
@@ -0,0 +1,70 @@
...
+index dab26bdea361..8f32802a1750 100644
+--- a/third_party/WebKit/Source/platform/wtf/typed_arrays/ArrayBufferContents.cpp
++++ b/third_party/WebKit/Source/platform/wtf/typed_arrays/ArrayBufferContents.cpp
+@@ -121,6 +121,11 @@ void* ArrayBufferContents::AllocateMemoryOrNull(size_t size,
+   return AllocateMemoryWithFlags(size, policy, base::PartitionAllocReturnNull);
+ }
+ 
++void* ArrayBufferContents::Realloc(void* data, size_t size) {
++  return Partitions::ArrayBufferPartition()->Realloc(data, size,
++    WTF_HEAP_PROFILER_TYPE_NAME(ArrayBufferContents));
```
alexeykuzmin: Indentation here seems to be a little bit off )

PR #626, opened 2018-07-31
src: add patches for Node upgrade v10.6.0

codebytere:
This PR adds patches for the Node upgrade to `v10.6.0`. It also corrects the `get-patch` script doc for the `--output-dir` cli flag.

/cc @alexeykuzmin


alexeykuzmin:
Some patches do not apply cleanly.

Btw, what `src: ` prefix means in the PR title? )

codebytere:
@alexeykuzmin ah sorry it's a Node convention whereby the prefix is the code directory being affected

PR #627, opened 2018-08-01
fix: Backport "Use unverified sync token in WebGL's DrawingBuffer."

jlfwong:
This is a backport of
https://chromium-review.googlesource.com/c/chromium/src/+/696666
targeting the 2-0-x branch of Electron.

This is a follow up to #624, which backported a commit which preceeded this one. It turns out that we need both in order to actually resolve the problem.


jlfwong:
We're going through local verification now to build confidence that the issue really is solved with this patch applied, and I'll write back here to update before we merge this.

jlfwong:
Okay, we've verified on two different laptops that this does indeed resolve our problem. Once this goes green, this should be good to merge & bump the associated version in electron/electron

PR #628, opened 2018-08-02
fix: backport legacy OCSP APIs in boringssl

nornagon:
One of Electron's tests exposed the fact that boringssl didn't support the OCSP stapling APIs that nodejs was calling. This backports a patch from boringssl that adds the APIs that nodejs expects, and causes the test to pass.

### DIFF
--- a/patches/common/boringssl/.patches.yaml
+++ b/patches/common/boringssl/.patches.yaml
@@ -1,0 +1,6 @@
+repo: src/third_party/boringssl/src
+patches:
+-
+  owners: nornagon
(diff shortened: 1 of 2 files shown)


nornagon:
I was mistaken—this doesn't actually fix the test. The real issue is that [BoringSSL doesn't automatically build certificate chains by default](https://bugs.chromium.org/p/boringssl/issues/detail?id=42).

nornagon:
NB, this actually doesn't enable node to use the OCSP stapling APIs, because it checks for the api with `#if defined(SSL_CTX_set_tlsext_status_cb)`, but BoringSSL defines that as a function, not a macro.

PR #633, opened 2018-08-06
fix: [mas] remove usage of private APIs

nornagon:
See electron/electron#13957


nornagon:
hrm, heads up, this patch causes some `gfx_unittests` to fail upstream: https://ci.chromium.org/p/chromium/builders/luci.chromium.try/mac_chromium_rel_ng/111640

Gonna see if i can figure out what's going on.

nornagon:
Looks like my assessment that `_CFIsObjC` always returns 0 was wrong. Going to have to find another workaround.

PR #643, opened 2018-08-09
chore: roll clang 307486:308728

nornagon:
See upstream:

https://chromium-review.googlesource.com/c/chromium/src/+/582074

We don't need any of the other changes from that CL, because they're all
fuchsia- or iOS-only.

Ref electron/electron#13988

### DIFF
--- a/patches/093-roll-clang.patch
+++ b/patches/093-roll-clang.patch
@@ -1,0 +1,21 @@
+diff --git a/tools/clang/scripts/update.py b/tools/clang/scripts/update.py
+index 8d9e1ebb56c4..6eaed04c4a22 100755
+--- a/tools/clang/scripts/update.py
++++ b/tools/clang/scripts/update.py
+@@ -27,14 +27,14 @@ import zipfile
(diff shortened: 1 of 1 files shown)


K900:
Well, this is unfortunate :(

The error seems really weird, too - it's like the system libstdc++ is actually somehow newer than what ships with this Clang roll?

deepak1556:
The error seems to be fixed in https://chromium-review.googlesource.com/c/chromium/src/+/582976 , so using `308728-3` should be good for us.

K900:
Nice. @nornagon you want to try that roll with libcc before I try it with Electron itself?

nornagon:
Updated the patch to refer to `308728-3`.

nornagon:
I'll wait on @deepak1556 to merge this when ready.

K900:
Does this need to go in first, and then the submodule on the Electron PR needs to be updated?

nornagon:
Yep, that's accurate.

K900:
Then I guess we need to make some sort of a final call on whether this is worth merging, both here and on Electron itself.

PR #647, opened 2018-08-14
build: optionally commit patches after they have been applied

alexeykuzmin:
Adds an optional `--commit` arg to the `apply-patches` script.
Shouldn't affect regular builds in any way, but will allow the GN build to commit libcc patches after they have been applied to prevent issues with untracked files in the Chromium tree.

Step 2 of this change will be converting all `.patch` files to a format that [`git-format-patch`](https://git-scm.com/docs/git-format-patch) understands, this way we will get rid of authors and descriptions in the `.patches.yaml` files and will be able to commit those patches easier.
It will be done in a separate PR.

/cc @nornagon

### DIFF
--- a/patches/common/angle/.patches.yaml
+++ b/patches/common/angle/.patches.yaml
@@ -1,6 +1,6 @@
 repo: src/third_party/angle
 patches:
 -
-  owners: alespergl
+  author: Ales Pergl <alpergl@microsoft.com>
   file: dcheck.patch
   description: null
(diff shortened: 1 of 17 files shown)


alexeykuzmin:
@jkleinsc Does packages src help to speed up the builds?
I'm trying to commit libcc patches here, but since the package src is used, it seems like patching fails since `src` is not a git repo.

alexeykuzmin:
@ckerr fixed

PR #649, opened 2018-08-15
fix: add patch to re-enable osr MouseWheelPhaseHandler

codebytere:
Backport [this commit](https://bitbucket.org/chromiumembedded/cef/commits/fad6aec5d0f02853ff6b8af403c5b75d4764e3cb) from CEF in order to add `MouseWheelPhaseHandler` in osr. 

Related: https://github.com/electron/electron/pull/14074

/cc @brenca @alexeykuzmin @deepak1556


deepak1556:
The patch seems to be from a relatively newer chromium version, can you try creating the patch from libcc master ?

PR #653, opened 2018-08-16
build: Add optional --project-root arg for the patching related scripts

alexeykuzmin:
It is necessary because we have different folders hierarchy
in traditional build setup with libchromiumcontent being a separate repo
and in gn-based build with libchromiumcontent repo checked out inside the src/.

Usage:
`./script/apply-patches --project-root=SRC_PARENT_FOLDER`

##### Checklist

### DIFF
--- a/script/patch.py
+++ b/script/patch.py
@@ -11,6 +11,7 @@ import os
 import subprocess
 import sys
 
+import lib.git as git
 from lib.patches import Patch, PatchesList, PatchesConfig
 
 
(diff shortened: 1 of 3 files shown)


script/patch.py
```diff
@@ -62,6 +64,8 @@ def parse_args():
   parser = argparse.ArgumentParser(description='Apply patches to a git repo')
   parser.add_argument('-f', '--force', default=False, action='store_true',
                       help='Do not stop on the first failed patch.')
+  parser.add_argument('--project-root', required=False, default=git.get_repo_root(os.path.abspath(__file__)),
+                      help='Parent folder to resolve repos relative paths against')
   parser.add_argument('-R', '--reverse', default=False, action='store_true', help='Apply patches in reverse.')
   parser.add_argument('-r', '--repo', help='Path to a repository root folder.')
```
deepak1556: Seems like `--repo` can serve the same purpose as `--project-root` in this script, can we remove one in favor of the other ?

alexeykuzmin: They are mutually exclusive, so technically yes.
But they have different semantics, so no =)

`--project-root` is used to explicitly set a parent folder for the `src/`, while `--repo` is used when a path to single patch is passed.

PR #657, opened 2018-08-17
build: ignore comparison differences between res files

ZacWalk:
##### Description of Change

Build compiles rc files twice once with rc.exe and another with rc.py. Ignoring comparison of output in the build due to a bug in rc.py

##### Checklist

### DIFF
--- a/patches/common/chromium/ignore_rc_check.patch
+++ b/patches/common/chromium/ignore_rc_check.patch
@@ -1,0 +1,17 @@
+diff --git a/build/toolchain/win/tool_wrapper.py b/build/toolchain/win/tool_wrapper.py
+index a76e926a8681..c43839a01211 100644
+--- a/build/toolchain/win/tool_wrapper.py
++++ b/build/toolchain/win/tool_wrapper.py
+@@ -258,7 +258,11 @@ class WinTool(object):
+       if rc_exe_exit_code == 0:
+         import filecmp
+         # Strip "/fo" prefix.
+-        assert filecmp.cmp(rc_res_output[3:], rcpy_res_output[3:])
++        # ------
++        # Temporarily ignore compares
++        # Nightly builds use very large version numbers that fail this check
++        # FIXME(zacwalk): Enable the assert.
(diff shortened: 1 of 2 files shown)


patches/common/chromium/ignore_rc_check.patch
```diff
@@ -0,0 +1,16 @@
...
++++ b/build/toolchain/win/tool_wrapper.py
+@@ -258,7 +258,10 @@ class WinTool(object):
+       if rc_exe_exit_code == 0:
+         import filecmp
+         # Strip "/fo" prefix.
+-        assert filecmp.cmp(rc_res_output[3:], rcpy_res_output[3:])
++        # ------
++        # Temporarily ignore compares         # 
++        # Nightly builds use very large version numbers that fail this check
++        # assert filecmp.cmp(rc_res_output[3:], rcpy_res_output[3:])
```
ckerr: Having seen several comments that say "Temporarily"... please add a `FIXME(ZacWalk)` to this

alexeykuzmin: Maybe to it would be better to add a comment in the `.patches.yaml`?

patches/common/chromium/ignore_rc_check.patch
```diff
@@ -0,0 +1,16 @@
...
+index a76e926a8681..c43839a01211 100644
+--- a/build/toolchain/win/tool_wrapper.py
++++ b/build/toolchain/win/tool_wrapper.py
+@@ -258,7 +258,10 @@ class WinTool(object):
+       if rc_exe_exit_code == 0:
+         import filecmp
+         # Strip "/fo" prefix.
+-        assert filecmp.cmp(rc_res_output[3:], rcpy_res_output[3:])
++        # ------
++        # Temporarily ignore compares         #
```
alexeykuzmin: Looks like we have some training nonsense at this line?

patches/common/chromium/ignore_rc_check.patch
alexeykuzmin: @ckerr Done.

PR #658, opened 2018-08-21
perf: commit our patches in batches instead of 1 by 1

MarshallOfSound:
This updates @alexeykuzmin's commit logic to commit our patches in batches so that we don't run commit a lot of times on a massive repo.  This drops the re-sync time from 10 minutes to 2 minutes on my local machine, disabling commits drops it to 1min 52 seconds so this is wayyy better now 👍

### DIFF
--- a/script/lib/git.py
+++ b/script/lib/git.py
@@ -78,21 +78,53 @@ def get_head_commit(repo):
     return subprocess.check_output(args).strip()
 
 
-def commit(repo, author, message):
+commits = {}
+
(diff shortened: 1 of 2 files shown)


MarshallOfSound:
> So it would be great to come up with some compromise solution that wouldn't create unnecessary complications for the development workflow.

Putting this change behind a flag that's enabled by default would probably be the best solution.  Most people will be 👌 with the faster builds and those like need to `git blame` inside chromium can do it by syncing with a different flag

PR #662, opened 2018-08-26
build: disable a DCHECK in viz::ServerSharedBitmapManager

alexeykuzmin:
It gets constantly triggered on Windows CI, e.g.
https://windows-ci.electronjs.org/project/AppVeyor/electron-i8wjp/build/1.0.91
We expect that backporting of https://chromium-review.googlesource.com/802574
should fix it.

##### Checklist

### DIFF
--- a/patches/common/chromium/.patches.yaml
+++ b/patches/common/chromium/.patches.yaml
@@ -34,6 +34,7 @@ patches:
       base/memory/weak_ptr.cc
       base/process/kill_win.cc
       components/viz/service/display/program_binding.h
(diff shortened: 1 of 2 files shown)


deepak1556:
Just adding a note for reviewers, the backport is planned after ch67 is merged. Also the failure only happens on ci, running the testing builds locally behind `--disable-gpu --disable-gpu-compositing --disable-early-init` doesn't trigger the crash, we are unable to isolate the reason for this, so the plan is to backport that CL which fixes some ownership of `ServerSharedBitmapManager` class.

PR #673, opened 2018-09-05
fix: workaround for X11ErrorTrackers issue

ckerr:
##### Description of Change

Fixes the crashing powerMonitor test that was disabled @ https://github.com/electron/electron/commit/76f26341bf3bc79f748fb2725251ee56fe7f37ab, so that now it can be re-enabled safely.

Only one X11ErrorTracker should exist at a time, but upstream has a bug where two can exist if running in headless mode --

1. ui::(anonymous namespace)::SupportsEWMH() [inner tracker is created]
1. ui::WmSupportsHint()
1. ui::IsX11WindowFullScreen()
1. ui::ScreensaverWindowFinder::IsScreensaverWindow()
1. ui::ScreensaverWindowFinder::ShouldStopIterating()
1. ui::EnumerateTopLevelWindows()
1. ui::ScreensaverWindowFinder::ScreensaverWindowExists() [outer tracker created]
1. ui::CheckIdleStateIsLocked()
1. ui::CalculateIdleState()

Removal of either tracker could have side-effects in some code paths, so ultimately this is probably handled better by a refactor that is out-of-scope for Electron. Instead, this patch removes the check that prevents more than one tracker from existing at a time.

cc @deepak1556 @MarshallOfSound 

##### Checklist


codebytere:
@ckerr doesn't look like this patch applies cleanly, exits on `script/update`

codebytere:
@ckerr this also needs to be PR'd into https://github.com/electron/libchromiumcontent/tree/electron-3-0-x

ckerr:
The problem with removing either `X11ErrorTracker` is that both `ScreensaverWindowExists()` and `SupportsEWMH()` base their return values on information from the trackers. If we remove one, one of those functions is going to start returning incorrect values in some conditions.

Enforcing the existence of only one `X11ErrorTracker at a time seems to be an architectural safeguard: if an outer and inner function are both looking at low-level XErrors, then that code needs to be refactored. At the code level, though, it doesn't seem to do harm: when the inner `X11ErrorTracker` goes out of scope, it will restore the previous error handler, similar to how nested `errno` checkers operate. That's why this is the safer workaround.

I completely agree that either approach is a workaround and that the real fix is to not have this code path to begin with. Before reporting this upstream, I want to confirm that it's valid to call `ui::CalculateIdleState()` when headless. If it's not, then the bug is ours and we could change the Atom code that calls `ui::CalculateIdleState()`.

patches/common/chromium/.patches.yaml
```diff
@@ -456,4 +475,4 @@ patches:
     to fix the order of V8 and Blink initialization.  Also fixes order
     of DidCreateScriptContext notification with initialization of window globals
     to fix electron/electron#13787.  The backport landed in Chromium 67 but the
-    DidCreateScriptContext re-ordering needs to be upstreamed or kept indefinitely
+    DidCreateScriptContext re-ordering needs to be upstreamed or kept indefinitely
```
alexeykuzmin: The newline at the end shouldn't have been removed =/

patches/common/chromium/allow_nested_error_trackers.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/ui/gfx/x/x11_error_tracker.cc b/ui/gfx/x/x11_error_tracker.cc
+index af031de356c5..2a5c18dc473a 100644
+--- a/ui/gfx/x/x11_error_tracker.cc
++++ b/ui/gfx/x/x11_error_tracker.cc
+@@ -24,7 +24,7 @@ namespace gfx {
+ X11ErrorTracker::X11ErrorTracker() {
+   // This is a non-exhaustive check for incorrect usage. It disallows nested
+   // X11ErrorTracker instances on the same thread.
+-  DCHECK(g_handler == NULL);
++  // DCHECK(g_handler == NULL);
```
alexeykuzmin: This is scary.

PR #676, opened 2018-09-07
fix: do not let Chromium manage webview's frame's lifetime

zcbenz:
##### Description of Change

This is part of fixes for https://github.com/electron/electron/issues/14211.

As explained in https://github.com/electron/electron/issues/14211#issuecomment-418616406, Chromium has some race conditions when managing WebFrame of webview that, it will sometimes accidentally detach the WebFrame of webview, resulting in the WebContents of webview being wrongly destroyed.

This patch disables Chromium's ability to detach the WebFrame of webview. There will be another patch in Electron to manually manage the WebContents of webview on Electron's side.

##### Checklist


codebytere:
@zcbenz this also needs to be backported into the `electron-3-0-x` branch, i can do that this morning

PR #678, opened 2018-09-07
fix: do not let Chromium manage webview's frame's lifetime (3-0-x)

codebytere:
##### Description of Change

Backport https://github.com/electron/libchromiumcontent/pull/677 to `3-0-x`.

/cc @alexeykuzmin @deepak1556


codebytere:
already done, missed it!

PR #682, opened 2018-09-10
fix: [m67] patch leveldb to not redefine ssize_t

nornagon:
##### Description of Change
node.js (and its dependencies) define the `ssize_t` type in their headers on Windows, using a guard to prevent redefinition. leveldb also defines the `ssize_t` type in the `port_chromium.h` file, but without the redefinition guard. This causes a compilation error on windows, because the type is defined twice when compiling `atom_api_browser_window.cc` in electron.

##### Checklist

### DIFF
--- a/patches/common/chromium/.patches.yaml
+++ b/patches/common/chromium/.patches.yaml
@@ -458,3 +458,10 @@ patches:
     This is part of the fixes for https://github.com/electron/electron/issues/14211.
     We should revisit this bug after upgrading to newer versions of Chrome,
     this patch was introduced in Chrome 66.
+-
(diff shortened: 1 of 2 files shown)


alexeykuzmin:
Is there a chance to upstream it?

nornagon:
Yes, I think this is potentially upstreamable.

deepak1556:
Just saw this, reminded of the hack we did during ch63 upgrade i guess for the exact issue https://github.com/electron/electron/blob/master/atom/browser/api/atom_api_web_contents.cc#L10-L14, I think that hack can be removed safely in 67 now with this patch in place. @nornagon can you remove it in the 67 branch ? Thanks!

PR #696, opened 2018-10-24
fix: proxy resolution from pac file for ws requests on windows and macOS (3-0-x)

deepak1556:
##### Description of Change

Backports https://chromium-review.googlesource.com/c/chromium/src/+/1143093
Fixes https://github.com/electron/electron/issues/15344

This is not required for versions greater than 3.x because we changed to use chromium proxy resolver implementation with https://github.com/electron/electron/pull/14656

And also it is not required for versions less than 3.x as the switch to platform proxy resolver was made temporarily only for ch 66 upgrade due to lack of support for mojo network context apis.

##### Checklist

### DIFF
--- a/patches/common/chromium/.patches.yaml
+++ b/patches/common/chromium/.patches.yaml
@@ -412,3 +412,10 @@ patches:
 
     This is required by the nativeWindowOpen option, which put multiple main
     frames in one renderer process.
+-
+  author: deepak1556 <hop2deep@gmail.com>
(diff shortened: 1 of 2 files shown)


MarshallOfSound (approved):

alexeykuzmin (approved):

ckerr:
@deepak1556 thanks for the description of why this isn't needed except for in 3.x :+1:

PR #705, opened 2018-11-22
build: ship mojo proxy resolver (3-1-x)

deepak1556:
##### Checklist


felixrieseberg:
Thank you for the backport. The change looks okay to me - I know that @nornagon won't be crazy excited about the additional patch, but I don't see another way to get this building.

felixrieseberg (approved):

deepak1556:
The patch is same the one from original PR, only additional change is the exposure of interfaces for accessing the proxy service from content layer which is necessary with gyp build.

PR #740, opened 2018-12-17
fix: Magellan sqlite security issue

ckerr:
##### Description of Change

As per the meeting discussion this morning, this PR backs out the larger sqlite 3 version bump and instead cherry-picks the upstream Magellan fixes from https://www.sqlite.org/src/vinfo/d44318f59044162e to avoid schema changes that users can't downgrade out of.

Note to reviewers: the `fts3ScanInteriorNode()` corruption tests that trigger `rc = FTS_CORRUPT_VTAB; goto finish_scan;` differed between upstream's pre-patch 3.26 and amalgamation's pre-patch 3.0. I've kept both tests here to be safe.

CC @deepak1556 @nornagon 

##### Checklist

- [ ] Patch information is added to appropriate `.patches.yaml`


nornagon (approved):
LGTM

PR #690, opened 2018-09-25
fix: expose libgtkui::GetBgColor (backport: 3-0-x)

codebytere:
##### Description of Change

Backport libcc portion of https://github.com/electron/electron/pull/14785 to `3-0-x`.

/cc @alexeykuzmin 

##### Checklist

- [x] PR description included and stakeholders cc'd
- [x] Patch information is added to appropriate `.patches.yaml`
- [ ] `script/update` runs without error
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

PR #742, opened 2018-12-31
fix: expose libgtkui::GetFgColor (backport: 3-1-x)

Cobinja:
##### Description of Change
Backport libcc portion of https://github.com/electron/electron/pull/15878 to `3-1-x`
/cc @inukshuk

##### Checklist

##### Notes

`srcipt/update`runs into the (probably) same error as in https://github.com/electron/libchromiumcontent/pull/690#issuecomment-424504022
The patch file actually gets applied, but the script tells otherwise

### DIFF
--- a/patches/common/chromium/libgtkui_export.patch
+++ b/patches/common/chromium/libgtkui_export.patch
@@ -10,6 +10,15 @@ index 665ec57..4ccb088 100644
  
  namespace aura {
  class Window;
+@@ -177,7 +177,7 @@ void ApplyCssToContext(GtkStyleContext* context, const std::string& css);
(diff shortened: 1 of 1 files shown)


codebytere:
hey @Cobinja this patch doesn't apply due to mismatched line numbers; you'll need to ensure it applies locally with `script/update`. We can get this merged once that's fixed up!

Cobinja:
@codebytere I forced-pushed a new version. It wasn't the line numbers you suspected. I can't tell what it actually was but I suspect a confusion of line-ending types

codebytere (approved):

PR #743, opened 2019-01-23
fix: don't call SetInitialBounds on macOS

zcbenz:
Don't call SetInitialBounds on macOS, since the initial size has already been set by BridgedNativeWidget, and the SetInitialBounds of Widget has bug that will set wrong size.

Fix https://github.com/electron/electron/issues/16015.


nornagon (approved):
I'd like to add a test for this but given it's fixed in 4 I'm not too worried

PR #749, opened 2019-03-06
fix: network hangs when using DHCP based WPAD (backport: 3-1-x)

deepak1556:
##### Description of Change

Backports:
https://chromium-review.googlesource.com/c/chromium/src/+/876921/
https://chromium-review.googlesource.com/c/chromium/src/+/946870/

Upstream Issue:
https://bugs.chromium.org/p/chromium/issues/detail?id=770201

This is not required to be backported to 3-0-x, since 3-0-x uses the platforms proxy resolver which is the equivalent of using `--proxy-resolver-winhttp` flag in chrome. And as mentioned in https://bugs.chromium.org/p/chromium/issues/detail?id=770201#c61 users in that config are not affected.

##### Checklist

### DIFF
--- a/patches/common/chromium/.patches.yaml
+++ b/patches/common/chromium/.patches.yaml
@@ -483,3 +483,16 @@ patches:
   description: |
     Update sandbox rules that target 10.13 exactly to target 10.13 and higher.
     Backports https://chromium-review.googlesource.com/c/chromium/src/+/1200184/
+-
+  owners: deepak1556
+  file: add_netlog_details_for_dhcp-based_wpad.patch
+  description: |
+    Adds NetLog instrumentation for how the network adapter list was read on Windows.
+    Backports https://chromium-review.googlesource.com/c/chromium/src/+/876921/
+-
+  owners: deepak1556
+  file: skip_network_adapters.patch
+  description: |
+    Skip network adapters that are not in state IfOperStatusUp when probing for
+    WPAD via DHCP.
+    Backports https://chromium-review.googlesource.com/c/chromium/src/+/946870/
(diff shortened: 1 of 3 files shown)


nornagon (approved):
Are these straightforward backports? Were there any significant conflicts to resolve?

deepak1556:
Its a direct backport of those CLs, manual intervention was not required.

deepak1556:
Forgot to squash and merge, sorry!

bpasero:
@deepak1556 hey, we are getting increased reports of users that connection of VSCode onto network drives (via UNC path) is slow: https://github.com/Microsoft/vscode/issues/70045. People say this was not happening with 3.1.2. 

We did change Electron from 3.1.2 to 3.1.6. Can you recall if any change (such as this one) could have an impact?

bpasero:
I will conduct a test with our users to use 3.1.2 again with our 1.32 release. Since there is people also complaining on Linux, the root of this issue might be elsewhere. Will report back when I get feedback on how it goes with the previously used 3.1.2

deepak1556:
@bpasero I am pretty sure its not caused by this patch. Did vscode change how it loads workspace directories ? There is an upstream issue that seems to reflect the issue described above https://chromium-review.googlesource.com/c/chromium/src/+/1234335/ , the fix got only into 70. If I can get a net-log trace from the users, I can confirm if its the same issue and backport the fix to 3.1.x line.

bpasero:
@deepak1556 just heard back from users that this reproduces with an older Electron version, so it seems unrelated. We after all just use node.js fs module to read files. I will continue to investigate on the VSCode side. Thanks for the link.

PR #750, opened 2019-03-07
fix: (FileReader) Make a copy of the ArrayBuffer when returning partial result (backport: 2-0-x)

deepak1556:
##### Description of Change

Backports https://chromium-review.googlesource.com/c/chromium/src/+/1492873

##### Checklist


ppontes:
Don't you need to add the new patch file to .patches?

nornagon (approved):

